PR #12, opened 2018-01-30
Address issues #3 and #8

### LINKED ISSUE #3: Impact of INT on Path MTU
As transit INT devices expand the size of data packets by adding INT metadata to the packets, end hosts have to be able to discover path MTU accounting for INT. This aspect is not well addressed in the current specification. 

A desirable way of handling this would be for a INT transit hop at which packet size after adding INT metadata exceeds outgoing link MTU, to send an ICMP Destination Unreachable message with code = “fragmentation needed and DF set”, and an appropriately calculated MTU reported in the ICMP message, allowing end hosts to discover the right path MTU with INT being performed in the network.
 
The path MTU aspect needs to be clarified and expanded upon in the dataplane specification.

### LINKED ISSUE #8: Reserved INT instruction bits: clarify behavior for forward compatibility
There are currently 8 reserved INT instruction bits. We need to specify required behavior when receiving a packet with one or more of the reserved INT instruction bits set. The typical reserved field behavior, setting to zero on transmission and ignoring received values, is not sufficient since any future use of these bits would break backwards compatibility. If software versions are mixed in the same network, nodes running the older version will add less INT metadata than nodes running the newer version. This will make it difficult for the monitor to determine which INT metadata came from which node. It is also likely that there will be a mismatch between the amount of INT metadata added and the value of the instruction count, leading to an incorrect increment of the length in the IP header, and UDP header if present.

In order to allow the reserved bits to be used in future versions with backward compatibility, we need to check whether any of the reserved INT instruction bits is set. At first glance there seem to be two possible approaches:
1. A node receiving a packet with one or more of the reserved INT instruction bits set must not add any metadata to that INT hop-by-hop header.
2. A node receiving a packet with one or more of the reserved INT instruction bits set must add the corresponding metadata filled with "0xFF" in each octet.

Any node adding an INT hop-by-hop header must set all reserved instruction bits to 0.

mhira1:
Edits to INT.mdk to elaborate on on MTU handling (fixes #3) and handling of receiving reserved bits set at a transit hop (fixes #8)

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -43,7 +43,7 @@ Tex Header:
 
 [TOC]
 
-# Introduction 
+# Introduction
 
 Inband Network Telemetry (“INT”) is a framework designed to allow the
 collection and reporting of network state, by the data plane, without requiring
@@ -99,14 +99,14 @@ of Little Minions paper [^Minions].
 types of INT Headers -- *Hop-by-hop* and *Destination* (See Section
 [#sec-int-header-types]).
 
-**INT Packet**: Any packet containing an INT Header. 
+**INT Packet**: Any packet containing an INT Header.
 
 **INT Instruction**: Embedded packet instructions indicating which INT Metadata
 to collect (defined below). The collected data is written into an INT Header.
 
 **INT Source**: A trusted entity that creates and inserts INT Headers into the
 packets it sends. The INT Headers contain, at minimum, INT Instructions
-indicating what to collect. 
+indicating what to collect.
 
 **INT Sink**: A trusted entity that extracts the INT Headers and collects the
 path state contained in the INT Headers. The INT Sink is responsible for
@@ -126,7 +126,7 @@ information using the INT approach. In practice, however, it seems useful to
 define a small baseline set of metadata that can be made available on a wide
 variety of devices:  the metadata listed in this section comprises such a set.
 As the INT specification evolves and the INT technology becomes more popular, we
-expect to add more metadata to this INT specification. 
+expect to add more metadata to this INT specification.
 
 Note the exact meaning of the following metadata (e.g., the unit of timestamp
 values, the precise definition of hop latency or congestion status) can vary by
@@ -163,7 +163,7 @@ beyond the scope of this document. For example, some devices may encode the
 physical port ID, whereas the other may encode the logical port ID. Some other
 types of devices can encode both (2B each).
 * Hop latency
-  - Time taken for the INT packet to be switched within the device. 
+  - Time taken for the INT packet to be switched within the device.
 * Egress port TX Link utilization
   - Current utilization of the egress port via which the INT packet was sent
 out. Again, devices can use different mechanisms to keep track of the current
@@ -180,7 +180,7 @@ INT packet observes in the device while being forwarded.
   - The fraction (in percentage or decimal fraction) of current queue occupancy
 relative to the queue-size limit. This indicates how much buffer space was used
 relative to the maximum buffer space (instantaneously or statically) available
-to the queue. 
+to the queue.
 
 # Processing INT Headers
 
@@ -228,7 +228,7 @@ path (e.g., either a copy of the original INT packet or a digest of it). Upon
 receiving the trigger, the slow path should process the INT Header
 appropriately, generate a new packet -- called a “follow-up packet” --
 containing the execution results of the INT instructions. The follow-up packet
-is forwarded separately. 
+is forwarded separately.
 
 If devices in the network do perform slow-path INT processing, it is possible
 that a single INT packet could spawn multiple follow-up packets – and in turn
@@ -265,7 +265,7 @@ header provides sufficient space to carry the INT information and that both the
 INT Sources and Sinks can agree on the location of the INT Headers.  The
 following choices are all potential encapsulations using common protocol stacks,
 although the INT user may choose a different encapsulation format if better
-suited to their needs and environment. 
+suited to their needs and environment.
 
 * INT over VXLAN (as a VXLAN payload, per GPE extension)
 * INT over Geneve (as a Geneve option)
@@ -286,7 +286,7 @@ approach doesn’t rely on any tunneling/virtualization mechanism and is
 versatile to apply INT to both native and virtualized traffic.
 2. *INT over VXLAN* - Our examples use the VXLAN generic protocol extensions
 (draft-ietf-nvo3-vxlan-gpe) to carry INT Headers between VXLAN header and
-encapsulated VXLAN payload. 
+encapsulated VXLAN payload.
 3. *INT over Geneve* - Geneve is an extensible tunneling framework, allowing
 Geneve options to be defined for INT Headers.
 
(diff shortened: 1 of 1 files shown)


telemetry/specs/INT.mdk
```diff
@@ -298,18 +298,64 @@ exhausting header space in the case of a forwarding loop or any other anomalies,
...
+Alternatively, each INT hop may report an MTU only accounting for the metadata
+it inserts. This would enable the path MTU discovery source converge to a
+precise path MTU, at the cost of receiving more ICMP messages, one from each
+INT hop.
+
+Regardless of whether or not an INT transit switch participates in Path MTU
+discovery, if it cannot insert all requested metadata because doing so will
+cause the packet length to exceed egress link MTU, it
+**must not insert any metadata and set the M bit** in the
+INT header, indicating that egress MTU was exceeded at an INT hop.
```
jklr, line 334: Though we say below at the M bit description, let's clarify here that the device must not add any metadata to the packet.

telemetry/specs/INT.mdk
```diff
@@ -833,7 +905,7 @@ INT Shim Header for VXLAN-GPE, destination INT type:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
-INT Metadata Header and Metadata Stack, 
+INT Metadata Header and Metadata Stack,
 followed by encapsulated Ethernet payload:
 `
     0                   1                   2                   3
```
jklr, line 913: Please revise the header examples with the newly added M bit.

telemetry/specs/INT.mdk
```diff
@@ -983,30 +1055,30 @@ header int_header_t {
     bit<4>  instruction_mask_0811;
     bit<4>  instruction_mask_1215;
     bit<16> rsvd2;
-}   
-    
+}
```
jklr, line 1061: M bit in the int_header_t definition.

telemetry/specs/INT.mdk
```diff
@@ -1380,7 +1452,7 @@ control Int_egress(inout headers hdr, inout metadata meta,
 
     apply{
         if(hdr.int_header.isValid()) {
-            if(hdr.int_header.max_hop_cnt != hdr.int_header.total_hop_cnt 
+            if(hdr.int_header.max_hop_cnt != hdr.int_header.total_hop_cnt
```
jklr, line 1459: We can add a condition that checks MTU limit and an action that sets the M bit.

telemetry/specs/INT.mdk
```diff
@@ -298,18 +298,64 @@ exhausting header space in the case of a forwarding loop or any other anomalies,
...
+(Instruction Count\*4\*INT Hop Count + 8), based on conservative values of
+total number of INT hops and instruction count being used, will prevent egress
+MTU being exceeded due to INT metadata insertion at INT hops.
+
+* An INT source/transit switch may optionally aid in dynamic discovery of Path MTU
+along flows being monitored by INT by transmitting ICMP message as per Path MTU
+Discovery mechanisms of the corresponding IP protocol
+(RFC 1191 for IPv4, RFC 1981 for IPv6). An INT source or transit switch may report
+a conservative MTU in the ICMP message, accounting for metadata insertion at
+downstream INT hops and assuming uniform egress MTU at downstream hops. This
```
mickeyspiegel: Not just downstream hops. All hops including upstream and downstream should be subtracted from the underlying real MTU of the egress link.

telemetry/specs/INT.mdk
mhira1: It is already clarified both here and below in the M bit description.

telemetry/specs/INT.mdk
mhira1: Done.

telemetry/specs/INT.mdk
mhira1: Done.

telemetry/specs/INT.mdk
mhira1: Int_egress currently does not have access to egress MTU. Can you please modify the P4 program to pass in egress MTU to Int_egress and set the M bit if needed in Int_egress?

telemetry/specs/INT.mdk
```diff
@@ -298,18 +298,64 @@ exhausting header space in the case of a forwarding loop or any other anomalies,
...
+
+An INT source needs to be able to insert 8 bytes of INT header plus Instruction Count \* 4
+bytes of its own metadata. If inserting 8 bytes of INT header itself causes
+egress link MTU to be exceeded, INT will not be initiated for such
+packets (**TBD - handling of such scenarios**). If there is enough room to
+insert INT header, but not first-hop INT metadata at the source, the source must
+initiate INT and set the M bit at the first hop itself.
+
+In theory, an INT transit switch can perform IPv4 fragmentation to overcome
+egress MTU limitation when inserting its metadata. However, IPv4 fragementation
```
mickeyspiegel: typo: fragementation --> fragmentation

telemetry/specs/INT.mdk
```diff
@@ -298,18 +298,64 @@ exhausting header space in the case of a forwarding loop or any other anomalies,
...
+can have adverse impact on applications. Moreover, IPv6 packets cannot be
+fragmented at intermediate hops. Also, fragmenting packets at INT transit hops,
+with or without copying preceding INT metadata into fragments imposes
+extra complexity of corelating fragments in the INT monitoring engine.
```
mickeyspiegel: typo: corelating --> correlating

mhira1:
Incorporated all comments, except for handling egress MTU and M bit in the P4 program. @jklr can you please make this change to the P4 program?

telemetry/specs/INT.mdk
```diff
@@ -612,14 +649,49 @@ metadata to the packet. If Total Hop Count (see below) equals Max Hop Count, a
...
+header.
+  - When adding a new metadata, each device **must prepend** its metadata in
+front of the metadata that are already added by the upstream devices. This is
+similar to the push operation on a stack. Hence, the most recently added
+metadata appears at the top of the stack. The device must add metadata in the
+order of bits set in the instruction bitmap.
+  - If a device is unable to provide a metadata value specified in the
+instruction bitmap because its value is not available, it must add a special
+reserved value  0xFFFF_FFFF indicating “invalid”.
+  - If a device cannot add all the number of bytes (i.e., Instruction Count\*4) required by the instruction bitmap (irrespective of the availability of the
```
mickeyspiegel: Break the long line in two?

mickeyspiegel (approved):

telemetry/specs/INT.mdk
mhira1: Agreed, fixed this.

telemetry/specs/INT.mdk
mhira1: Done

telemetry/specs/INT.mdk
mhira1: Done

telemetry/specs/INT.mdk
mhira1: Done

PR #18, opened 2018-02-14
Add probe marker option to indicate the existence of INT in IP TCP/UDP packets

heidinet:
As discussed at P4 application meeting on Jan 18,  a new proposal was presented to satisfy certain deployment scenarios, where DSCP field can't be assigned for INT,  and TCP/UDP ports need to be kept to retain ECMP behavior.   With this proposal, probe marker with arbitrary values are inserted after TCP/UDP header as an indicator of INT existence. This probe marker was originally proposed by IETF draft https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01 (expired), and has been implemented by ASIC vendors.

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -414,8 +414,8 @@ retaining ECMP behavior is desired, probe marker option could be followed.
 With arbitrary values being inserted after TCP/UDP header in probe marker
 option, the likelihood of conflicting with user traffic in a data center is 
 low, but cannot be completely eliminated. To further reduce the chance of 
-conflict, a user deploys probe marker option could choose to check more into
-port number to identify the existence of INT.) 
+conflict, the user who deploys probe marker option could choose to check more 
+into TCP/UDP port numbers to identify the existence of INT.) 
 
 
 INT over TCP/UDP adds INT metadata after TCP/UDP headers as if the payload is
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -404,7 +404,12 @@ the reserved or configured values are
 per domain decisions that all INT capable devices should follow. (Note: in
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
-ACL and network debugging.)
+ACL and network debugging. If DSCP field can not be reseved for INT, and 
+retaining ECMP behavior is desired, adding probe marker fields could be an
+option. With arbitary values being inserted after TCP/UDP header, the chance
+to conflict with user traffic in a data center is low, but can not be completely 
+eliminated.)
+
 
 INT over TCP/UDP adds INT metadata after TCP/UDP headers as if the payload is
 changed. However, the sink device will remove INT headers before passing the
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -371,7 +371,7 @@ both INT header stacks carry information for respective layers and need not be
 considered interfering with each other.)
 
 A field in the lower layers of Ethernet, IP, or TCP/UDP should indicate if the
-INT header exists after the TCP/UDP header. We propose two options:
+INT header exists after the TCP/UDP header. We propose three options:
 
 * IPv4 DSCP or IPv6 Traffic Class field: A value or a bit will be reserved to
 indicate the existence of INT after TCP/UDP. INT source will put the reserved
@@ -384,8 +384,23 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 32-bit values are inserted after IP TCP/UDP header
+to indicate the existence of INT after TCP/UDP. These fileds should be interpreted
+as unsigned integer values and are initialized to a configured value. 
 
-The decision to use DSCP or destination port field and the reserved values are
+INT probe marker for TCP/UDP:
+
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |                         Probe Marker (1)                      |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |                         Probe Marker (2)                      |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+
+The decision to use DSCP, destination port field or probe marker, and 
+the reserved or configured values are
 per domain decisions that all INT capable devices should follow. (Note: in
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -388,8 +388,9 @@ Port, and sink must restore the original port number.
 header to indicate the existence of INT after TCP/UDP. These fields should be 
 interpreted as unsigned integer values, stored in network byte order and are 
 initialized to a configured value.  This Probe marker is a variation of an early 
-IETF draft with existing implementations. More information can be found at 
-https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01.
+IETF draft with existing implementations [^DPP].
+[^DPP] : Data-plane probe for in-band telemetry collection,
+https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01
 
 INT probe marker for TCP/UDP:
 
@@ -415,7 +416,7 @@ With arbitrary values being inserted after TCP/UDP header in probe marker
 option, the likelihood of conflicting with user traffic in a data center is 
 low, but cannot be completely eliminated. To further reduce the chance of 
 conflict, the user who deploys probe marker option could choose to check more 
-into TCP/UDP port numbers to identify the existence of INT.) 
+into TCP/UDP port numbers to validate INT probe marker) 
 
 
 INT over TCP/UDP adds INT metadata after TCP/UDP headers as if the payload is


telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,32 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 32-bit values are inserted after IP TCP/UDP header
```
jklr: I think you mean 64-bit values?
"after TCP/UDP header, before INT Shim header"

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,32 @@ ignore the designated bit position.)
...
+the reserved or configured values are
 per domain decisions that all INT capable devices should follow. (Note: in
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
-ACL and network debugging.)
+ACL and network debugging. If DSCP field can not be reseved for INT, and 
+retaining ECMP behavior is desired, adding probe marker fields could be an
+option. With arbitary values being inserted after TCP/UDP header, the chance
+to conflict with user traffic in a data center is low, but can not be completely 
+eliminated.)
```
jklr: Do you want to mention the optional check on TCP/UDP port numbers?

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,32 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 32-bit values are inserted after IP TCP/UDP header
+to indicate the existence of INT after TCP/UDP. These fileds should be interpreted
+as unsigned integer values and are initialized to a configured value.
```
jklr: Can we say that the value is also per-domain decision?

telemetry/specs/INT.mdk
jklr: And how about citing the IETF draft of DPP?

telemetry/specs/INT.mdk
masoudmj: What's the purpose of mentioning "unsigned integer values"?

telemetry/specs/INT.mdk
masoudmj: can not -> cannot

telemetry/specs/INT.mdk
heidinet: I just added the IETF draft links, also corrected the bits

telemetry/specs/INT.mdk
heidinet: The "per-domain" was mentioned in below context.
For the "integer" just added context "These fileds should be interpreted
as unsigned integer values, stored in network byte order"

telemetry/specs/INT.mdk
heidinet: Added the port check and also corrected to "cannot"

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,36 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 64-bit values are inserted after IP TCP/UDP header
+to indicate the existence of INT after TCP/UDP. These fileds should be interpreted
```
mhira1: Typo. fileds=fields

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,36 @@ ignore the designated bit position.)
...
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
-ACL and network debugging.)
+ACL and network debugging. If DSCP field cannot be reseved for INT, and 
+retaining ECMP behavior is desired, adding probe marker fields could be an
+option. With arbitary values being inserted after TCP/UDP header, the chance
+to conflict with user traffic in a data center is low, but cannot be completely 
+eliminated. A user could choose to combine port number along with 64-bit marker
+as an INT indicator to reduce the conflicting even more.) 
+
```
mhira1: Perhaps leave out the last sentence. If a destination port number is dedicated to INT, what is the use-case to add 64-bit marker also? If the concern is probe marker conflicting with real data, then why would such a deployment not use destination port alone? Opening up possibility of combining destination port with probe marker will mean that all switches will need to understand probe markers. You cannot have a switch that only recognizes INT destination port as it will start processing information after TCP header as INT headers, which will actually be a probe marker.

mhira1 (approved):

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,36 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 64-bit values are inserted after IP TCP/UDP header
+to indicate the existence of INT after TCP/UDP. These fileds should be interpreted
+as unsigned integer values, stored in network byte order and are initialized to a 
+configured value.  This Probe marker is a variation of an early IETF draft
+https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01 (expired), with existing
+implemetations.
```
mhira1: We can leave out reference to an expired draft. This approach is justified even without support from the draft.

mhira1 (dismissed):

telemetry/specs/INT.mdk
heidinet: corrected it.

telemetry/specs/INT.mdk
heidinet: It's mainly for pointing it out that there is implementation based on that.

telemetry/specs/INT.mdk
heidinet: This extra "port" check applies to only "probe marker" option, shouldn't impact non-probe-marker approach. I will re-word it.

mhira1: Thanks for clarifying. I  misunderstood this to mean a combination of destination port number approach to identifying INT and probe marker approach to identify INT. Please reword. JK/I will add clarification regarding combinations of approaches.

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,39 @@ ignore the designated bit position.)
...
+It is recommended that only one option is deployed per domain. (Note: in
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
-ACL and network debugging.)
+ACL and network debugging. If DSCP field cannot be reserved for INT, and 
+retaining ECMP behavior is desired, probe marker option could be followed.
+With arbitrary values being inserted after TCP/UDP header in probe marker
+option, the likelihood of conflicting with user traffic in a data center is 
+low, but cannot be completely eliminated. To further reduce the chance of 
+conflict, a user deploys probe marker option could choose to check more into
```
masoudmj: the user who deploys probe marker 
or
the user deploying probe marker

It is not clear what is the "port number" referring to. You may want to change it to:
check more into the TCP/UDP destination port number

mhira1 (commented):
Regarding the last clause, thanks for clarifying that the destination port is not INT L4 destination port.

Even then, I am wondering whether there is a need for combining probe marker + L4 destination port. If one is doing INT only on probe packets, then why use L4 destination port + probe marker to identify INT, why not just L4 destination port. If one is doing INT on real application, then rewriting L4 dest port will cause traffic to take a different ECMP path.

I agree with allowing for probe marker approach as an alternative when DSCP space is not available, but am wondering whether combination of L4 dest port + probe marker is needed.

masoudmj:
@mhira1 I believe the L4 destination port is combined with probe marker as a conservative measure. For example, If after the TCP/UDP comes the application data, it may collide with the chosen probe marker but if we only apply probe marker to the specific protocol on top of TCP/UDP we know that after UDP it comes some header with specific format that never collides with probe marker so we are safe.

jklr:
Right, the port number matching is a safety measure, not a way to actively identify INT.
There will be a *set* of port numbers that the probe marker are applied to. Hence this limits the usage of probe-marker-based-INT to those predefined set of applications. From the pkts belonging to the set of applications, probe marker is used to differentiate INT encapped pkts.

telemetry/specs/INT.mdk
heidinet: Changed it to "the user who deploys probe marker option could choose to check more
into TCP/UDP port numbers to identify the existence of INT."  I want to keep the flexibility to check either source or destination port.

masoudmj (approved):

mhira1 (approved):

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,39 @@ ignore the designated bit position.)
...
 general, encoding into DSCP field will be less intrusive compared to changing
 the layer 4 port field. The latter may alter ECMP behavior and can complicate
-ACL and network debugging.)
+ACL and network debugging. If DSCP field cannot be reserved for INT, and 
+retaining ECMP behavior is desired, probe marker option could be followed.
+With arbitrary values being inserted after TCP/UDP header in probe marker
+option, the likelihood of conflicting with user traffic in a data center is 
+low, but cannot be completely eliminated. To further reduce the chance of 
+conflict, the user who deploys probe marker option could choose to check more 
+into TCP/UDP port numbers to identify the existence of INT.)
```
jklr: how about this clarification? 
"the existence of INT" => "the existence of INT probe marker"

telemetry/specs/INT.mdk
```diff
@@ -384,12 +384,39 @@ ignore the designated bit position.)
 * TCP/UDP destination Port field: A port number in layer 4 will be reserved to
 indicate the existence of INT after TCP/UDP. INT source changes the destination
 Port, and sink must restore the original port number.
+* New Probe Marker fields: arbitrary 64-bit values are inserted after IP TCP/UDP 
+header to indicate the existence of INT after TCP/UDP. These fields should be 
+interpreted as unsigned integer values, stored in network byte order and are 
+initialized to a configured value.  This Probe marker is a variation of an early 
+IETF draft with existing implementations. More information can be found at 
+https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01.
```
jklr: Please change the last two sentences as follows:

> This Probe marker is a variation of an early IETF draft with existing implementations [^DPP].
> [^DPP] :  Data-plane probe for in-band telemetry collection,  https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01

jklr (approved):

telemetry/specs/INT.mdk
heidinet: I am adding the 2 [^DPP], assume it's the pdf format.

telemetry/specs/INT.mdk
heidinet: Make it more restrict:
to identify the existence of INT   => "to validate INT probe marker"

PR #21, opened 2018-03-01
Support for two levels of port identifiers

### LINKED ISSUE #2: Reporting physical and logical port ID in telemetry metadata
Certain use-cases would require reporting of logical port IDs (e.g. tunnel ID), while certain use-cases may require reporting of both logical and physical port IDs.

Address how logical port IDs may be differentiated from physical port IDs, and how both logical and physical port IDs may be reported in certain use-cases. Existing 4 byte field may not suffice for reporting both logical and physical port IDs.

mhira1:
This commit enhances the INT dataplane specification to
enable  monitoring of two levels of ingress and egress
port identifiers.

Fixes #2


mhira1:
Will create a new pull request.

PR #25, opened 2018-03-14
Added Checksum Complement

talmi:


telemetry/specs/INT.mdk
```diff
@@ -377,6 +377,44 @@ Considering all these factors, the INT specification requires that an INT
...
+The motivation for the Checksum Complement is that some hardware implementations
+process data packets in a serial order, which may impose a problem when INT
+fields and metadata that reside after the L4 Checksum field are inserted or 
+modified. Therefore, the Checksum Complement metadata, if present, is the last
+metadata field in the stack. 
+
+Note that when the Checksum Complement metadata is present source/transit 
+switches may choose to update the L4 Checksum field instead of using the
+Checksum Complement metadata. In this case the Checksum Complement will be 
+assigned zero. A host that verifies the L4 Checksum will be unaffected by
```
mhira1: In general, this spec uses 0xFFFFFFFF as the reserved value for metadata corresponding to an instruction that an INT hop does not process. Using a consistent reserved value would simplify implementations. An implementation can simply insert this single reserved value for any bits in the instruction bit mask that it does not support, without understanding and distinguishing between the bits that it does not support.

telemetry/specs/INT.mdk
```diff
@@ -377,6 +377,44 @@ Considering all these factors, the INT specification requires that an INT
...
+* Assign a value to the Checksum Complement metadata which guarantees that the 
+existing L4 Checksum is the correct value of the packet after the INT-related 
+updates.
+
+The motivation for the Checksum Complement is that some hardware implementations
+process data packets in a serial order, which may impose a problem when INT
+fields and metadata that reside after the L4 Checksum field are inserted or 
+modified. Therefore, the Checksum Complement metadata, if present, is the last
+metadata field in the stack. 
+
```
mhira1, line 417: We should add some clarification here that the INT source indicates that checksum neutral updates are allowed by setting an instruction bit corresponding to checksum complement metadata. The above text on INT transit being able to perform L4 checksum update in one of two ways seems to indicate that this is always the case. If the INT source does not set the checksum complement instruction bit, then transit hops have to update the L4 checksum (except if L4 checksum is zero for VXLAN or GENEVE in which case checksum update can be skipped altogether).

mhira1 (commented):
Tal, thanks for making these edits. Please see comments inline. Let's discuss further at tomorrow's working group meeting.

talmi:
The comments above have been addressed in the latest commit.
Thanks,
Tal.

mhira1 (approved):

jklr (approved):

PR #26, opened 2018-03-14
Add telemetry opaque data to report format

mickeyspiegel:
This allows implementations to experiment, adding arbitrary data to telemetry reports in between the telemetry report header and the packet snippet in the payload.


mickeyspiegel:
There does not seem to be any appetite for defining ranges for the 24-bit Schema IDs of the Opaque state snapshot field in IOAM.
The proposal here is to define ranges of the 16-bit Schema ID in the Telemetry extensions header in telemetry reports. Most of the values are for private or experimental use, i.e. do whatever you want with them. These values are put at the bottom of the range, since it is more likely that IOAM usages will involve Schema IDs at the bottom of the range. This maximizes the chances for the same Schema ID value to be used in IOAM and in telemetry reports. At the top, a range of 255 values is reserved for specification by p4.org. Another range was carved out just in case there is any desire to do FCFS allocation in the future, though this range could easily be removed.

telemetry/specs/telemetry_report.mdk
```diff
@@ -398,6 +399,45 @@ particular deployment.
...
+      |                                                               |
+      |                        Extensions data                        |
+      |                                                               |
+      |                                                               |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+`
+
+* NProt: Next Protocol
+  : Same definition as **NProt** in the Telemetry Report Header defined in
+    Section [#telem-report-header].
```
jklr, line 420: do we allow a chain of extension headers?

telemetry/specs/telemetry_report.mdk
```diff
@@ -398,6 +399,45 @@ particular deployment.
...
+
+* Length
+  : Indicates the length of the telemetry extensions header in multiples
+    of 4 octets.
```
jklr, line 424: If I'm correct, the length of this extension hdr is NOT added to the length of the main report hdr. Shall we clarify that, maybe at the definition of the main hdr length field?

telemetry/specs/telemetry_report.mdk
```diff
@@ -398,6 +399,45 @@ particular deployment.
   : The device local time when the packet was first received on the ingress
     physical or logical port, in nanoseconds.
 
+## Telemetry Extensions Header
```
jklr, line 402: There are multiple 'extension' hdrs, e.g., the domain-specific extension in INT spec.
How about calling this one as "Report Extensions" header?
'Report' seems the main keyword that distinguishes this extension from other extensions.

jklr (commented):
The context around IETF/IOAM extension hdr seems useful and relevant. They're currently available on in the PR. Wonder if we want to move some of them into the spec. The details may change also in IETF but maybe we can capture the intention in the spec. Thanks.

jklr:
@mickeyspiegel can you pls respond to the comments? so that we can merge it. Thx!

mickeyspiegel:
Superseded by #61

PR #29, opened 2018-03-24
Added "Layer 4 Checksum Update" flag in INT header

mhira1:
Added support for allowing Layer 4 checksum update to be
optional at transit hops, in scenarios where the INT
source knows that the destination does not validate
Layer 4 checksum.

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -1,6 +1,6 @@
 Title         : In-band Network Telemetry (INT)
 Title Note : (NOTE: This is a working draft. Consider using tagged versions for implementations.)
-Title Footer: 2018-03-14
+Title Footer: 2018-03-26
 Author : The P4.org Applications Working Group
 Affiliation : Contributions from *Alibaba, Arista, Barefoot Networks, Dell, Intel, Marvell, VMware*
 Heading depth: 5
@@ -123,7 +123,7 @@ inserts into the INT Header.  Examples of metadata are described below.
 The INT devices within the same domain must be configured in a consistent way to
 ensure interoperability between the devices.
 Operators of an INT domain should deploy INT Sink capability at domain edges to
-prevent INT information from leaking out of the domain. 
+prevent INT information from leaking out of the domain.
 
 # What To Monitor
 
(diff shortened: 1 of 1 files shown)


telemetry/specs/INT.mdk
```diff
@@ -384,44 +384,71 @@ Considering all these factors, the INT specification requires that an INT
...
+be set with caution at an INT source that is enabling INT on probe packets,
+or at an INT source that is also a Tunnel Endpoint and knows that the
+receiving Tunnel Endpoint does not validate outer Layer 4 checksum.
+
+If an INT transit hop receives UDP packets with checksum equal to zero,
+or a packet with this "Checksum update optional" bit set in the INT header,
+it may choose to not update the Layer 4 checksum upon inserting INT metadata.
+If an INT transit hop does update the Layer 4 checksum in such packets, it
+**must** perform a full checksum calculation as the original checksum in the
+packet cannot be assumed to be valid for an incremental checksum calculation.
```
mickeyspiegel: I see two different cases:

1. If it receives UDP packets with checksum equal to zero, then I don't think the INT spec is the place to define how to change the checksum. IETF RFCs specify how to behave when receiving a packet with UDP checksum equal to zero. IMO this is out of scope of INT and does not need to be discussed in this spec.

2. If the "checksum update optional" bit is set, then it does not matter what the INT transit hop does. The resulting checksum must be ignored at the receiving tunnel endpoint. There is no need for this requirement.

telemetry/specs/INT.mdk
```diff
@@ -384,45 +384,95 @@ Considering all these factors, the INT specification requires that an INT
...
+are removed at the INT sink, making the original Layer 4 checksum valid after
+the INT sink.
+* If INT is enabled on real data packets and INT source is also the Layer 4
+traffic source, it may compute the Layer 4 checksum on the entire Layer 4
+payload including INT headers. In this case, if INT sink is also the Layer 4
+destination, every INT transit hop must update the Layer 4 checksum.
+If the INT sink is not the Layer 4 destination, transit hops may not update
+the checksum (unless INT metadata integrity is to be protected), the
+INT sink must update the Layer 4 checksum upon removing INT headers and
+metadata, before forwarding the data packet to the destination.
```
mickeyspiegel, line 429: This paragraph is difficult to decipher. I have tried multiple times and cannot tell what it is saying, or what requirements it is imposing.

Why does it matter whether the INT sink is the L4 destination or not?

Language such as "transit hops may not update the checksum" is confusing. It should be clear from the received packet whether checksum integrity must be preserved, or the checksum must be left unaltered. It is not up to the transit hop to decide, is it?

telemetry/specs/INT.mdk
```diff
@@ -384,45 +384,95 @@ Considering all these factors, the INT specification requires that an INT
...
+
+This specification defines a bit in the INT header that is set at the INT
+source to dictate whether or not subsequent INT hops must update the
+Layer 4 checksum.
+
+If an INT transit/sink switch receives UDP packets with checksum equal to zero,
+it must honor the intention at the UDP source and maintain zero UDP checksum
+in the packet as per RFC 768 and RFC 6936. If an INT transit/sink switch
```
mickeyspiegel, line 437: The first sentence is also true for INT source, if it is not the traffic source or tunnel source.

telemetry/specs/INT.mdk
```diff
@@ -384,45 +384,95 @@ Considering all these factors, the INT specification requires that an INT
...
-* Update the L4 Checksum field such that the new value is equal to the Checksum 
+encapsulation header that includes an L4 header, such as VXLAN.
+
+In theory, since each INT transit hop modifies the Layer 4 payload, it should
+update the Layer 4 checksum. Layer 4 checksum update at transit hops is
+necessary to protect the integrity of INT metadata between the INT source
+and INT sink. If integrity of INT metadata is to be protected by Layer 4
+checksum, each INT transit hop must perform incremental checksum calculation
+as complete checksum calculation at a transit hop will mask any corruption
+in the INT metadata inserted by previous INT hops.
```
mickeyspiegel, line 398: Since this paragraph is so early in this section, it reads as if protecting the integrity of INT metadata is the key concern when determining whether to maintain checksum integrity when adding/removing INT metadata. I don't believe that is the case. If this justification is to remain, it should be discussed after the UDP checksum zero and encapsulation protocol aspects, and integrity of the L4 payload.

telemetry/specs/INT.mdk
```diff
@@ -685,7 +735,7 @@ INT Metadata Header and Metadata Stack:
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5     6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |  Ver  |Rep|C|E|M|     Reserved      | Hop ML  |RemainingHopCnt|
+   |  Ver  |Rep|D|E|M|C|    Reserve      | Hop ML  |RemainingHopCnt|
```
mickeyspiegel, line 738: Reserved

mickeyspiegel (changes requested):
The latest rewrite goes into a fair amount of depth on potential reasons to maintain checksum integrity. This makes the whole section much harder to follow. It would be good to streamline the text a bit further.

mhira1:
Abandoning the proposal for these edits, keeping things simple for version 1.0. We will only skip checksum update if UDP checksum is zero. Otherwise every INT hop adding/removing metadata must update the Layer 4 checksum. Will send a simpler set of edits to this effect separately.

We can revisit the need of an INT header flag for defining checksum update behavior post 1.0 based on use cases.

PR #30, opened 2018-03-26
minor corrections and editorial changes

jklr:
- Added egress timestamp under Egress Information section to be consistent with Ingress Information section 
- Corrected the size of fixed INT headers, considering shim/option/tail headers


telemetry/specs/INT.mdk
```diff
@@ -342,9 +346,12 @@ This may be addressed in the following ways -
 * It is recommended that the MTU of links between INT sources and sinks is
 configured to be higher than the MTU of preceding link MTUs (server/VM NIC MTUs)
 by an appropriate amount. Configuring an MTU differential of
-(Per-hop Metadata Length\*4\*INT Hop Count + 8), based on conservative values of
+[Per-hop Metadata Length\*4\*INT Hop Count + 12] bytes, based on conservative values of
```
mhira1: Thanks for catching the error. Thinking more, even Metadata + 12 bytes is not enough. We have to look at the maximum overhead added by INT across different encapsulation types. For TCP/IP, the overhead is Per-hop ML*4*Number of Hops + 16 (4 bytes shim header + 8 bytes INT header + 4 bytes Tail header).

telemetry/specs/INT.mdk
```diff
@@ -366,9 +373,9 @@ cause the packet length to exceed egress link MTU, it
 **must not insert any metadata and set the M bit** in the
 INT header, indicating that egress MTU was exceeded at an INT hop.
 
-An INT source needs to be able to insert 8 bytes of INT header plus Per-hop
-Metadata Length\*4
-bytes of its own metadata. If inserting 8 bytes of INT header itself causes
+An INT source needs to be able to insert 12 bytes of fixed INT shim header and
```
mhira1: This number is 16 for INT over TCP/IP. 12 for VxLAN-GPE, 12 for Geneve.

Also, we need to clarify that this is "additional" INT overhead for which MTU needs to be tweaked, not accounting for MTU adjustments for VxLAN and Geneve itself.

telemetry/specs/INT.mdk
```diff
@@ -195,8 +193,13 @@ physical or logical port.
     The semantics of port identifiers may differ across devices, each INT hop
     chooses the port type it reports at each of the two levels.
 
+* Egress timestamp
+  - The device local time when the INT packet was processed by the egress
+physical or logical port.
```
mickeyspiegel, line 198: I like this definition better than the one in Appendix A.3. Can we replace the definition in Appendix A.3 with this one?

telemetry/specs/INT.mdk
```diff
@@ -397,17 +404,20 @@ receiver to ignore the value of the Checksum field. For UDP over IPv6, there
...
+of the new packet, after the INT-related updates (header additions/removals, 
+field updates), or
 
 * If the INT source indicates that Checksum-neutral updates are allowed by 
 setting an instruction bit corresponding to the Checksum Complement metadata, 
 then the INT source/transit switches may assign a value to the Checksum 
 Complement metadata which guarantees that the existing L4 Checksum is the 
-correct value of the packet after the INT-related updates.
+correct value of the packet after the INT-related updates. INT sink doesn't
+need to take any special action but just removing the Checksum complemenet
```
mhira1: We need to clarify that sink has to update the L4 checksum, it cannot just remove the checksum complement. Sink is removing bytes from the packet, hence needs to recompute checksum.

This is also related to my pull request on the checksum-ignored flag. If you don't mind, revert your changes in the section on checksum. I can take care of these edits in my PR.

telemetry/specs/INT.mdk
mickeyspiegel: This is better, but it is still ignoring the tail header over TCP/UDP

telemetry/specs/INT.mdk
```diff
@@ -397,17 +404,20 @@ receiver to ignore the value of the Checksum field. For UDP over IPv6, there
 are specific use cases in which it is possible to assign a zero Checksum 
 (as defined in RFC 6936).
 
-If an L4 Checksum update is required, an INT source/transit switch may perform
+If an L4 Checksum update is required, an INT source/transit/sink switch may perform
```
mickeyspiegel: This does not work. The sink cannot use checksum complement, nor can it assume that removal of checksum complement fields is enough. If an L4 checksum update at the sink is required, it must update the L4 Checksum field.

mickeyspiegel (changes requested):
We still need to resolve the checksum issue

telemetry/specs/INT.mdk
mhira1: Another point to clarify is that this overhead assumes that only Per-hop INT metadata header is in use. No data is being piggybacked using Src-to-Sink INT metadata header.

telemetry/specs/INT.mdk
mhira1: Correct, the magic number is 16

jklr:
Reverted most of the checksum related changes since there is another PR handling the checksum section.

telemetry/specs/INT.mdk
jklr: Done.

telemetry/specs/INT.mdk
jklr: Done.

telemetry/specs/INT.mdk
jklr: Done. I called out the total Fixed INT Header Length as another variable and defined it per encap types.

mickeyspiegel (approved):

PR #31, opened 2018-03-27
Remove slow path follow-up packets and piggybacked metadata

mickeyspiegel:
Temporarily remove slow path processing using follow-up packets, and piggybacked metadata for closed loop control. Both could benefit from a better understanding of use cases and some preliminary implementation experience.


mickeyspiegel:
Further motivation:

Currently there are two INT Header Type values defined: Hop-by-Hop type (with value 1 in the examples) and Destination type (with value 2 in the examples). However, there are multiple incompatible uses of the Destination type mentioned as possibilities:

1. Add fields that are not of interest and should not be repeated at transit hops, such as sequence number and original IP TTL value. In this case the "Destination" Header must use a different format than the Hop-by-Hop Header.

2. Follow-up packets generated by slow path processing of INT. In this case the contents of the "Destination" header have the same format as the Hop-by-Hop Header.

3. Piggybacked INT hop-by-hop metadata, sent from the original INT Sink back to the original INT Source for purposes of closed loop control. Again in this case the contents of the "Destination" header have the same format as the Hop-by-Hop Header.

This points out that the details for processing of follow-up packets and piggybacked hop-by-hop metadata are not yet sufficiently mature to achieve interoperability.

Rather than suggest codepoint values that could be used in the examples, it would be better to specify code point values further up when they are necessary, and just have the examples follow those definitions.

mickeyspiegel:
The slow path processing is also unnecessarily complex. Why make an extra copy of each packet in the data plane at each hop, just to send each copy to the sink so that it can generate a report? Instead each hop can generate its own report using the postcard mode, as described in the telemetry report specification.

telemetry/specs/INT.mdk
```diff
@@ -222,68 +222,27 @@ There are two types of INT Headers, *hop-by-hop* and *destination*.  A given INT
...
-exact format of a follow-up packet other than its INT portion.
+The data format of Destination type INT Headers is not defined in this version,
+but is expected to be defined in future versions of this specification.
 
-To prevent potential packet ordering issues, it is recommended that an INT
-device **_NOT_** forward the INT packets themselves via the slow path while
-processing INT Headers.
+More INT Header types may be defined in future versions, for example to deliver
+piggybacked hop-by-hop metadata, from the original INT Sink back to the original
+INT Source for purposes of closed loop control.
```
mhira1: How is this different from destination type INT header?

mhira1 (commented):
Minor comment, otherwise ok.

telemetry/specs/INT.mdk
mickeyspiegel: The contents are different. For piggybacked metadata, the format is exactly the same as the INT Hop-by-Hop Metadata Header, it just needs an indication (proposing different INT type) that this is to be passed through and not added to each hop.
For destination type INT header with contents such as sequence number or original IP TTL, these are contents that differ from the contents of the INT Hop-by-Hop Metadata Header.

mhira1: If I understand and recall correctly, exact definition of src-to-sink INT message was considered outside the scope of the spec. All this spec needs to define is two types of INT headers: hop-by-hop type and destination type. The format of destination type INT message only needs to be understood by the INT source and sink and different implementations could use their own formats, in a way this is "opaque" data. The destination type INT message could carry piggybacked metadata, or it could carry sequence numbers, or original TTL or something totally different.

Perhaps, just define two types, type 1 (hop-by-hop) and type 2 (destination type) and leave out this one sentence about more INT header types?

mhira1:
Looks good.

PR #32, opened 2018-03-29
More precise definition of checksum behavior + Miscellaneous changes for v1.0

mhira1:


telemetry/specs/INT.mdk
```diff
@@ -731,7 +726,7 @@ INT Metadata Header and Metadata Stack:
   different number of INT hops.
 
 * The fields in the INT metadata header are interpreted the following way:
-  - Ver (4b): INT metadata header version. Should be zero for this version.
+  - Ver (4b): INT metadata header version. Should be 1 for this version.
```
jklr, line 726: Thanks for correcting this critical statement.

telemetry/specs/INT.mdk
```diff
@@ -955,7 +945,7 @@ INT Shim Header for TCP/UDP, INT type is hop-by-hop:
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |     Type=1    |   Reserved    |  Length = 8   |   Reserved    |
+   |     Type=1    |   Reserved    |  Length = 10  |   Reserved    |
```
jklr: length = 8 looks to be right. why is it 10?

jklr (approved):
looks good, only but the length field value change to 10.

telemetry/specs/INT.mdk
mhira1: Thanks for catching this. I started adding switch3 metadata, hence bumped up length to 10. But then realized that this switch3 is INT sink in this example, forgot to revert length back to 8.

Will change this back to 8.

PR #42, opened 2018-04-20
Some more editorial fixes for v1.0

mhira1:


telemetry/specs/INT.mdk
```diff
@@ -1252,7 +1258,7 @@ header tcp_t {
 header udp_t {
     bit<16> srcPort;
     bit<16> dstPort;
-    bit<16> length_;
+    bit<16> length;
```
mickeyspiegel: I thought there was a reason we should not call it length, but I am not sure. JK?

mickeyspiegel (approved):

jklr (approved):

PR #43, opened 2018-04-21
v1.0 spec PDFs

jklr:
- Addressing comments from Michael Orr
- create v1.0 spec PDF


mhira1 (approved):

telemetry/specs/INT.mdk
```diff
@@ -1462,6 +1465,7 @@ parser EgressParserImpl(packet_in packet,
     state parse_tcp {
         packet.extract(hdr.tcp);
         transition select(hdr.ipv4.dscp) {
+	    /* &&& is a concatenation operator in p4_16 */
```
jafingerhut: A couple of places in the P4_16 spec call `&&&` a mask operator.  Concatenation is denoted with the `++` operator in P4_16.  I would recommend replacing 'concatenation' with 'mask' in this comment.

jklr: Thanks. I actually used it for masking, and put a wrong comment.

PR #48, opened 2018-06-21
Add yang model for metadata semantics

mickeyspiegel:


mickeyspiegel:
We need to decide whether there should be a separate model just for metadata semantics between nodes and monitors, or whether the yang model should cover all of data plane telemetry including normal management and configuration.

In openconfig, they strongly advocate a separation of config and operational state. If the scope remains narrow, then we won't have any config state, only operational state, so it deviates from openconfig style to some extent.

telemetry/code/models/p4-dtel-metadata-semantics.yang
```diff
@@ -0,0 +1,326 @@
...
+    leaf size-units {
+      type identityref {
+        base METADATA_SIZE_UNITS;
+      }
+      description
+        "Units of size";
+    }
+
+    leaf cell-size {
+      when "../size-units = 'METADATA_CELLS'" {
```
chrispsommers, line 186: Shouldn't this 'when' be around the entire leaf? I'm not sure why you'd make the description conditional, especially when there is another description below.

mickeyspiegel: From RFC 6020:
   The "when" statement makes its parent data definition statement
   conditional.  The node defined by the parent data definition
   statement is only valid when the condition specified by the "when"
   statement is satisfied.

chrispsommers: Got it, thanks. My yang is rusty!

mhira1 (approved):
Mickey, this is a good initial version of the model. Please go ahead and commit it, other contributors can refine further. As we discussed at the working group meeting, please capture the open items from discussions in working group meetings - either as comments in the model or as a separate file in the repository, as you feel appropriate.

PR #49, opened 2018-06-21
INT Extensions for Domain Specific Metadata

rsivakolundu:
Added Domain Specific flag, Domain Specific Metadata bit to the Instruction bitmap and INT Metadata Header /w Domain Specific Metadata format.


telemetry/specs/INT.mdk
```diff
@@ -854,11 +859,14 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  - bit14: Domain Specfic Metadata
   - bit15: Checksum Complement
   - The remaining bits are reserved.
   Each instruction requests 4 bytes of metadata to be inserted at each hop,
-  except if bit 6 is set, which requires 8 bytes of metadata. Per-hop
-  metadata length is set accordingly at the INT source.
+  except if bit 6 and bit 14 is set. If bit 6 is set, the instruction requires 
+  8 bytes of metadata. If bit 14 is set, the instruction requires a domain specific
+  metadata of n bytes, n being a multiple of 4 bytes. Per-hop metadata length is 
+  set accordingly at the INT source.
```
mhira1, line 864: If only the source can set per-hop metadata length in accordance with the domain specific metadata, then a INT transit node should not be allowed to set the D bit as the per-hop metadata length will no longer be accurate.

telemetry/specs/INT.mdk
```diff
@@ -755,9 +755,9 @@ and the metadata itself.
 INT Metadata Header and Metadata Stack:
 `
     0                   1                   2                   3
-    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5     6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |  Ver  |Rep|C|E|M|     Reserved      | Hop ML  |RemainingHopCnt|
+   |  Ver  |Rep|C|E|M|D|    Reserved     | Hop ML  |RemainingHopCnt|
```
mhira1: Do we need the D bit if bit 14 in instruction bitmap conveys the same information?

mhira1 (commented):
Please see comments inline. Also, I am wondering if we can achieve this goal without growing the metadata header by 8 bytes. Let's discuss further at the next working group meeting.

telemetry/specs/INT.mdk
rsivakolundu: We discussed this bit in the WG meeting. I will add text to explain the use of this bit.

telemetry/specs/INT.mdk
rsivakolundu: We concluded that the bit is needed to convey the presence of Domain Specific Flags.

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+`
+
+* INT metadata header is 8 bytes long followed by a stack of INT metadata.
+  Each metadata is either 4 bytes or 8 bytes in length. Each INT hop adds
+  the same length of metadata. The total length of the metadata stack is
+  variable as different packets may traverse different paths and hence
+  different number of INT hops.
+
+* The fields in the INT metadata header are interpreted the following way:
+  - Ver (4b): INT metadata header version. Should be 1 for this version.
```
mhira1, line 943: Version should be 2 in this description

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+      limitation. In theory, if this does not occur at consecutive hops,
+      it may be possible for the monitoring system to derive which
+      switch(es) set the M bit based on knowledge of the network topology
+      and "Switch ID, Ingress port ID, Egress port ID" tuples in the INT
+      metadata stack.
+  - S (1b): Sink Node Specific Flag
+    - This flag is a INT Sink node specific flag to allow for the Sink nodes in the 
+      domain to process the metadata stack in a domain specific way.This flag must 
+      be set if a device, source node or transit node, wants to indicate to the sink
+      a domain specific handling of the metadata.
```
mhira1, line 992: My understanding is that we decided to decouple S bit processing from domain-specific processing, and define clearly what the sink is expected to do when the S bit is set. If you are still proposing S bit handling on a domain-specific basis, then there is some ambiguity that needs to be clarified in the spec -
1. S bit only makes sense when domain ID is non-zero, otherwise how do the node setting the S bit and sink node even know that they indeed support the same domain?
2. Does domain ID have to be set at INT source? Or can an intermediate node that is setting the S bit also set the domain ID?
3. If intermediate node can set the domain ID, how does it handle a packet in which domain ID is set at the source? Only set the S bit when the domain ID originally present in the packet matches its own domain ID?

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+  - bit2: Hop latency
+  - bit3: Queue ID (8 bits) + Queue occupancy (24 bits)
+  - bit4: Ingress timestamp
+  - bit5: Egress timestamp
+  - bit6: Level 2 Ingress Port ID + Egress Port ID (4 bytes each)
+  - bit7: Egress port Tx utilization
+  - bit14: Domain Specfic Instruction
+  - bit15: Checksum Complement
+  - The remaining bits are reserved.
+  Bits 0 - 13 are Baseline INT Instructions and Bit 14, Domain Specific Instruction.
```
mhira1, line 1030: Bits 0-13 and bit 15 are baseline INT instructions, and bit 14 indicates domain-specific instructions

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+values as specified in the instruction bitmap immediately after the INT metadata
+header.
+  - When adding a new metadata, each device must prepend its metadata in
+    front of the metadata that are already added by the upstream devices.
+    This is similar to the push operation on a stack. Hence, the most recently
+    added metadata appears at the top of the stack. The device must add
+    metadata in the order of bits set in the instruction bitmap.
+  - If a device is unable to provide a metadata value specified in the
+    instruction bitmap because its value is not available, it must add a special
+    all-ones reserved value indicating "invalid" (4 or 8 bytes of 0xFF
```
mhira1, line 1055: 4 or 8 bytes is not true any more, it may be N bytes of 0xFF if a node is skipping domain-specific metadata

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+    INT transit hop receives a reserved bit set in the instruction bitmap (e.g.
+    set by a INT source that is running a newer version), the transit hop must
+    either add corresponding metadata filled with the reserved value 0xFFFFFFFF
+    or must not add any INT metadata to the packet. This means that an
+    instruction bit marked reserved in this specification may be
+    used for a 4B metadata in a subsequent minor version while still being
+    backward compatible with this specification. However, an instruction bit
+    marked reserved in this specification may be used for a 8B metadata only
+    in the next major version, breaking backward compatibility and requring all
+    INT switches to be upgraded to the new major version. For example
```
mhira1, line 1072: In the version 2.0 description, this text should say that a 8B metadata may only be added in the next major version v3, example should talk about v2 v3 incompatibility vs v1, v2 incompatibility

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+    metadata stack will not be a multiple of Per-Hop Metadata length \* 4
+    in this case.
+  - If an INT transit hop does not add metadata to a packet due to any of the
+    above reasons, it must not decrement the remaining INT hop count in the INT
+    metadata header.
+* Summary of the field usage
+  - The INT Source must set the following fields:
+    - Ver, Rep, C, M, Per-hop Metadata Length, Remaining Hop Count,
+      and Instruction Bitmap.
+    - S bit may set by the source node if the INT metadata added by the source node meets
```
mhira1, line 1086: Again, I thought we were separating S bit from domain-specific handling, this needs further clarification

telemetry/specs/INT.mdk
```diff
@@ -906,35 +901,201 @@ header.
...
+  - bit3: Queue ID (8 bits) + Queue occupancy (24 bits)
+  - bit4: Ingress timestamp
+  - bit5: Egress timestamp
+  - bit6: Level 2 Ingress Port ID + Egress Port ID (4 bytes each)
+  - bit7: Egress port Tx utilization
+  - bit14: Domain Specfic Instruction
+  - bit15: Checksum Complement
+  - The remaining bits are reserved.
+  Bits 0 - 13 are Baseline INT Instructions and Bit 14, Domain Specific Instruction.
+  Domain Specific Instruction is an instruction that requires additional
```
mhira1, line 1031: Separate bullet point (* in mdk) for Domain Specific ID and domain specific flags. Also, we should make an explicit note here that addition of 4 octets of domain specific ID, domain specific flags breaks compatibility with version 1 as v1 INT switches will interpret this field as start of INT metadata stack

mhira1 (commented):
Thanks Ramesh for the changes. Apologize for not reviewing sooner. 

Please see comments inline. Also, let's add an explicit note where we start talking about v2 header format, that addition of new header fields in v2 breaks backward compatibility with v1.

telemetry/specs/INT.mdk
rsivakolundu: I will change it.

telemetry/specs/INT.mdk
rsivakolundu: S bit only makes sense when Domain ID is non-zero. I will add clarity.

The domain ID is set by the INT source only. If the domain ID in the packet doesn't match your domain ID, then the transit Node should not set the S bit.

telemetry/specs/INT.mdk
rsivakolundu: Agreed. I will change the text.

telemetry/specs/INT.mdk
rsivakolundu: Agreed. I will change the text.

telemetry/specs/INT.mdk
rsivakolundu: S bit handling is specific to the domain. I will add text to clarify the S bit handling

telemetry/specs/INT.mdk
rsivakolundu: I will  update the version compatibility text to address this.

rsivakolundu:
Will create a new pull request in a branch to manage multiple pull-requests separately for each issue.

PR #54, opened 2018-09-28
INT extensions for domain specific features

rsivakolundu:


telemetry/specs/INT.mdk
```diff
@@ -854,11 +854,14 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
   - bit5: Egress timestamp
   - bit6: Level 2 Ingress Port ID + Egress Port ID (4 bytes each)
   - bit7: Egress port Tx utilization
+  - bit14: Domain Specfic Metadata
```
mickeyspiegel: typo: Specfic --> Specific

telemetry/specs/INT.mdk
```diff
@@ -752,12 +752,12 @@ hop-by-hop INT header must fit in a single Geneve option.
 In this section, we define the format for INT hop-by-hop metadata headers,
 and the metadata itself.
 
-INT Metadata Header and Metadata Stack:
+INT Metadata Header and Metadata Stack (Version = 1):
```
jklr: Let's drop the entire section of Version 1.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
...
+INT Metadata Header and Metadata Stack (Version = 2):
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |Ver = 2|Rep|C|E|M|S|    Reserved     | Hop ML  |RemainingHopCnt|
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |      Instruction Bitmap       |        Domain Specific ID     |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |            DS Flags           |            Reserved           |
```
mickeyspiegel: If this is to incorporate the sink indication, then the last 16 bits should no longer be reserved. Either sink indication explicitly, or something less specific like "DS field".

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
 For INT over Geneve it is 8 bytes subtracted from (length in Geneve tunnel
 option header \* 4).
 
+
+INT Metadata Header and Metadata Stack (Version = 2):
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |Ver = 2|Rep|C|E|M|S|    Reserved     | Hop ML  |RemainingHopCnt|
```
mickeyspiegel: With "Domain Specific" expanding into "Reserved" (i.e. another 16 bits), the S bit is no longer needed in this first part of the header.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
...
+      limitation. In theory, if this does not occur at consecutive hops,
+      it may be possible for the monitoring system to derive which
+      switch(es) set the M bit based on knowledge of the network topology
+      and "Switch ID, Ingress port ID, Egress port ID" tuples in the INT
+      metadata stack.
+  - S (1b): Sink Node Specific Flag
+    - This flag is a INT Sink node specific flag to allow for the Sink nodes in the 
+      domain to process the metadata stack in a domain specific way.This flag must 
+      be set if a device, source node or transit node, wants to indicate to the sink
+      a domain specific handling of the metadata.
```
mickeyspiegel: Same comment as above. With "Domain Specific" expanding into "Reserved" (i.e. another 16 bits), the S bit is no longer needed in this first part of the header. There is plenty of room in the fields after the Domain Specific ID for these types of indications.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
...
+  - R (9b): Reserved bits.
+  - Hop ML (5b): Per-hop Metadata Length, the length of metadata, including the 
+    Domain Specific Metadata in 4-Byte words to be inserted at each INT hop.
+    - While the largest value of Per-hop Baseline Metadata Length is 31, the largest
+      value of Per-hop Domain Specific Metadata is variable and not specified.
```
mickeyspiegel: The latter part of this sentence does not make sense. The Hop ML includes per-hop Domain Specific Metadata, so the total including both Baseline and Domain Specific Metadata is limited to 31.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
...
+each bit corresponds to a specific standard metadata as specified in Section 3.
+  - bit0 (MSB): Switch ID
+  - bit1: Level 1 Ingress Port ID (16 bits) + Egress Port ID (16 bits)
+  - bit2: Hop latency
+  - bit3: Queue ID (8 bits) + Queue occupancy (24 bits)
+  - bit4: Ingress timestamp
+  - bit5: Egress timestamp
+  - bit6: Level 2 Ingress Port ID + Egress Port ID (4 bytes each)
+  - bit7: Egress port Tx utilization
+  - bit14: Domain Specfic Instruction
```
mickeyspiegel: typo: Specfic --> Specific

telemetry/specs/INT.mdk
```diff
@@ -910,6 +913,190 @@ from (shim header length \* 4).
...
+  - Reserved bits in the instruction bitmap are to be handled similarly. If an
+    INT transit hop receives a reserved bit set in the instruction bitmap (e.g.
+    set by a INT source that is running a newer version), the transit hop must
+    either add corresponding metadata filled with the reserved value 0xFFFFFFFF
+    or must not add any INT metadata to the packet. This means that an
+    instruction bit marked reserved in this specification may be
+    used for a 4B metadata in a subsequent minor version while still being
+    backward compatible with this specification. However, an instruction bit
+    marked reserved in this specification may be used for a 8B metadata only
+    in the next major version, breaking backward compatibility and requring all
```
mickeyspiegel: typo: requring --> requiring
This looks to be an old typo that is present in previous versions of this spec.

mickeyspiegel (changes requested):

telemetry/specs/INT.mdk
rsivakolundu: Corrected

telemetry/specs/INT.mdk
rsivakolundu: Corrected

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Agree. Corrected

telemetry/specs/INT.mdk
```diff
@@ -937,6 +1118,7 @@ port ID)
   - C = 0
   - E = 0 (Max Hop Count not exceeded)
   - M = 0 (MTU not exceeded at any switch)
+  - S = 0
```
mickeyspiegel: This change and all the changes below should be reverted as well since the S bit is removed.

telemetry/specs/INT.mdk
rsivakolundu: Corrected

telemetry/specs/INT.mdk
rsivakolundu: Agree.

telemetry/specs/INT.mdk
```diff
@@ -1003,7 +1185,7 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   | Ver=1 | 0 |0|0|0|     Reserved      | HopML=2 |RemainingHopC=6|
+   | Ver=1 | 0 |0|0|0|0|   Reserved      | HopML=2 |RemainingHopC=6|
```
mickeyspiegel: All these examples need the addition of DS ID, DS Flags, DS Field

telemetry/specs/INT.mdk
rsivakolundu: What if the baseline metadata and domain specific metadata exceeds 31 4-byte words? I guess no easy way to handle that. Let me think about it and we can discuss in the weekly meeting

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Should we add examples of Ver=2 instead?

mickeyspiegel: In version 2, absolutely!

telemetry/specs/INT.mdk
jklr: 31 4B words is a hard limit. Do u see a need to insert more than 124 bytes at one hop?

telemetry/specs/INT.mdk
jklr: We will do this PR only at version 2. Hence no need to worry about version 1 at all.

telemetry/specs/INT.mdk
rsivakolundu: Not in INT mode. Postcard mode can be utilized for that.

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Agree. Removed Ver=1

telemetry/specs/INT.mdk
```diff
@@ -992,11 +1010,25 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
   Details of the metadata semantics YANG model can be accessed at the link below:
   https://github.com/p4lang/p4-applications/blob/master/telemetry/code/models/p4-dtel-metadata-semantics.yang
   
-  Each instruction requests 4 bytes of metadata to be inserted at each hop,
-  except if bit 6 is set, which requires 8 bytes of metadata. Per-hop
-  metadata length is set accordingly at the INT source.
-  
+  Bits 0 - 13 are Baseline INT Instructions and Bit 14, Domain Specific Instruction.
```
jklr: "Bit 14 is Domain ..."

jklr (commented):
Do we want to leave some notes how it's related to the Domain Specific Extension of the report format?

telemetry/specs/INT.mdk
jklr: minor nit, let's correct.

telemetry/specs/INT.mdk
```diff
@@ -353,11 +355,11 @@ The exact meaning of the following metadata (e.g., the unit of timestamp
...
-resource limits, etc. Thus, defining the exact meaning of each metadata 
-is beyond the scope of this document. Instead we assume that the semantics 
-of metadata for each device model used in a deployment is communicated 
-with the entities interpreting/analyzing the reported data in an out-of-band 
-fashion.
+
+resource limits, etc. Thus, defining the exact meaning of each metadata is 
+beyond the scope of this document. Instead we assume that the semantics of 
+metadata for each device model used in a deployment is communicated with 
+the entities interpreting/analyzing thereported data in an out-of-band fashion.
```
mickeyspiegel: typo: "thereported" --> "the reported"

telemetry/specs/INT.mdk
```diff
@@ -441,6 +443,17 @@ Details of the metadata semantics YANG model can be accessed at the link below:
...
+  - The build-up of traffic in the buffer (in bytes, cells, or packets) that the
+INT packet observes in the device while being forwarded. Use case is when the buffer 
+is shared between multiple queues. The format of this 4-octet metadata field is switch 
+implementation specific and the metadata semantics YANG model shall describe the format 
+and units of this metadata field in the metadata stack.
+
+Details of the metadata semantics YANG model can be accessed at the link below:
+
+https://github.com/p4lang/p4-applications/blob/master/telemetry/code/models/p4-dtel-metadata-semantics.yang
+
```
mickeyspiegel: This change is a duplicate. Maybe a sync issue?

telemetry/specs/INT.mdk
```diff
@@ -883,10 +896,13 @@ INT Metadata Header and Metadata Stack:
...
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |Ver = 2|Rep|C|E|M|      Reserved     | Hop ML  |RemainingHopCnt|
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |Ver = 1|Rep|C|E|M|     Reserved      | Hop ML  |RemainingHopCnt|
+   |      Instruction Bitmap       |        Domain Specific ID     |
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |      Instruction Bitmap       |            Reserved           |
+   |         DS Instruction        |            DS Flags           |
```
mickeyspiegel: Nit: whitespace at the end of the line

telemetry/specs/INT.mdk
```diff
@@ -883,10 +896,13 @@ INT Metadata Header and Metadata Stack:
 `
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |Ver = 2|Rep|C|E|M|      Reserved     | Hop ML  |RemainingHopCnt|
```
mickeyspiegel: I guess we also need a change to the spec Title specifying that this is version 2.

telemetry/specs/INT.mdk
mickeyspiegel: Also bit 14 should be added to the list of bits above.

telemetry/specs/INT.mdk
```diff
@@ -992,11 +1009,25 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  of the Domain Specific Flags and Domain Specific Field is required and 
+  Domain Specific Metadata is appended to the Baseline Metadata before 
+  Checksum Complement is inserted.
+
+  Each instruction requests 4 bytes of metadata to be inserted at each hop,
+  except if bit 6 and bit 14 is set. If bit 6 is set, the instruction requires 
+  8 bytes of metadata. If bit 14 is set, the instruction requires a domain specific
+  metadata of n bytes, n being a multiple of 4 bytes. Per-hop metadata length is 
+  set accordingly at the INT source.
+
```
mickeyspiegel, line 1059: I find this whole section to be less clear than I would like. My attempt at reorganizing the text:

Each instruction requests 4 bytes of metadata to be inserted at each hop, except
for the following bits:

- When bit 6 is set, the instruction requests 8 bytes of metadata to be inserted at each hop.
- When the Domain Specific Instruction (bit 14) is set, this indicates that additional 
  processing of the Domain Specific ID, Domain Specific Flags (DS Flags) and 
  Domain Specific Field (DS Field) is required in order to determine what to insert.

  If the Domain Specific ID matches any Domain ID known to this node, after
  additional processing of the Domain Specific Flags and Domain Specific Field,
  Domain Specific Metadata is appended to the Baseline Metadata, before the
  Checksum Complement is inserted. The amount of Domain Specific Metadata
  must be a multiple of 4 bytes, determined from the other Domain Specific
  Fields and consistent with the per-hop metadata length (Hop ML) set by the
  INT source.

  If the Domain Specific ID does not match any Domain ID known to this node, then
  the node is required to either:
    - Pad the node's INT Metadata stack with the special all-ones reserved value for
      a Domain Specific Metadata length, calculated by subtracting from the Hop ML
      a length computed from all bits other than bit 14, or
    - Skip INT processing altogether and not insert any metadata into the packet.

telemetry/specs/INT.mdk
```diff
@@ -1141,9 +1174,11 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   | Ver=1 | 0 |0|0|0|     Reserved      | HopML=2 |RemainingHopC=6|
+   | Ver=2 | 0 |0|0|0|0|   Reserved      | HopML=2 |RemainingHopC=6|
```
mickeyspiegel: I don't see any definition of a new flag bit after the "M" bit.
The only change here should be "Ver=1" --> "Ver=2".

telemetry/specs/INT.mdk
```diff
@@ -1196,9 +1231,11 @@ INT Metadata Header and Metadata Stack:
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   | Ver=1 | 0 |0|0|0|     Reserved      | HopML=2 |RemainingHopC=5|
+   | Ver=2 | 0 |0|0|0|0|   Reserved      | HopML=2 |RemainingHopC=5|
```
mickeyspiegel, line 1264: I don't see any definition of a new flag bit after the "M" bit.
The only change here should be "Ver=1" --> "Ver=2".

telemetry/specs/INT.mdk
```diff
@@ -1253,9 +1290,11 @@ INT Metadata Header and Metadata Stack:
...
```
mickeyspiegel, line 1323: I don't see any definition of a new flag bit after the "M" bit.
The only change here should be "Ver=1" --> "Ver=2".

telemetry/specs/INT.mdk
```diff
@@ -992,11 +1009,25 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  INT Metadata stack with the special all-ones reserved value for Domain Specific 
+  Metadata length as calculated by subtracting the Baseline Metadata Length from Hop_ML.
   
+  If the Domain Specific ID matches the node's Domain ID then additional processing 
+  of the Domain Specific Flags and Domain Specific Field is required and 
+  Domain Specific Metadata is appended to the Baseline Metadata before 
+  Checksum Complement is inserted.
+
+  Each instruction requests 4 bytes of metadata to be inserted at each hop,
+  except if bit 6 and bit 14 is set. If bit 6 is set, the instruction requires
```
mickeyspiegel: s/"bit 6 and bit 14"/"bit 6 or bit 14"
or "bit 6 and/or bit 14"

mickeyspiegel (changes requested):
I see two major differences from the IETF's IOAM namespace beyond the additional fields:
1. When a transit node receives an INT packet with a Domain Specific ID that it does not recognize, is it still allowed to add its metadata to the packet?
2. Does the Domain Specific ID have any bearing on the semantics and/or metadata units? In the IETF this seems to be the main point of namespace.
It would be good to hash this out further, either aligning with the IETF or clarifying at least to ourselves if not in the spec, why should this be different?

telemetry/specs/INT.mdk
rsivakolundu: Removed Bit 14 and added text to address Domain Specific processing

telemetry/specs/INT.mdk
rsivakolundu: Fixed the typo.

telemetry/specs/INT.mdk
rsivakolundu: Deleted the duplicate.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Tried to clean up the text. Hope it is clearer now.

telemetry/specs/INT.mdk
rsivakolundu: What Flag bit after M are you referring to?

telemetry/specs/INT.mdk
rsivakolundu: Fixed it. Bit 14 is no longer needed.

telemetry/specs/INT.mdk
mickeyspiegel: You added whitespace at the end of the line. I thought it should be removed.

telemetry/specs/INT.mdk
```diff
@@ -1009,24 +1006,28 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  must be a multiple of 4 bytes, determined from the Domain Specific
+  Instruction and consistent with the per-hop metadata length (Hop ML) set by the
+  INT source.
 
   Each instruction requests 4 bytes of metadata to be inserted at each hop,
-  except if bit 6 and bit 14 is set. If bit 6 is set, the instruction requires 
-  8 bytes of metadata. If bit 14 is set, the instruction requires a domain specific
-  metadata of n bytes, n being a multiple of 4 bytes. Per-hop metadata length is 
-  set accordingly at the INT source.
+  except if bit 6. If bit 6 is set, the instruction requires
```
mickeyspiegel: nit: s/except if/except for

telemetry/specs/INT.mdk
```diff
@@ -1009,24 +1006,28 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  Instruction and consistent with the per-hop metadata length (Hop ML) set by the
+  INT source.
 
   Each instruction requests 4 bytes of metadata to be inserted at each hop,
-  except if bit 6 and bit 14 is set. If bit 6 is set, the instruction requires 
-  8 bytes of metadata. If bit 14 is set, the instruction requires a domain specific
-  metadata of n bytes, n being a multiple of 4 bytes. Per-hop metadata length is 
-  set accordingly at the INT source.
+  except if bit 6. If bit 6 is set, the instruction requires 
+  8 bytes of metadata. Per-hop metadata length is set accordingly at the INT source.
```
mickeyspiegel: This paragraph should be moved up to the Instruction bitmap section, in front of the Domain Specific fields.

telemetry/specs/INT.mdk
```diff
@@ -1009,24 +1006,28 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  * Domain Specific ID (16b): the unique ID of the INT Domain.
+  
+  * DS Instruction (16b): Instruction bit map specific to the INT domain identified by the 
+  Domain Specific ID. Domain Specific Instruction is an instruction that requires additional 
+  processing of Domain Specific Flags (DS Flags) for the INT Domain identified by Domain Specific ID. 
+  If the Domain Specific ID doesn't match the Domain ID of the node, then the transit node is 
+  required to pad the node's INT Metadata stack with the special all-ones reserved value for Domain Specific 
   Metadata length as calculated by subtracting the Baseline Metadata Length from Hop_ML.
   
   If the Domain Specific ID matches the node's Domain ID then additional processing
```
mickeyspiegel: I believe a node could have knowledge of more than one domain ID. That is why I suggested:
s/matches the node's Domain ID/matches any Domain ID known to this node

telemetry/specs/INT.mdk
```diff
@@ -1009,24 +1006,28 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
+  Bits 0 - 14 are Baseline INT Instructions.
+  
+  * Domain Specific ID (16b): the unique ID of the INT Domain.
+  
+  * DS Instruction (16b): Instruction bit map specific to the INT domain identified by the 
+  Domain Specific ID. Domain Specific Instruction is an instruction that requires additional 
+  processing of Domain Specific Flags (DS Flags) for the INT Domain identified by Domain Specific ID. 
+  If the Domain Specific ID doesn't match the Domain ID of the node, then the transit node is 
+  required to pad the node's INT Metadata stack with the special all-ones reserved value for Domain Specific 
   Metadata length as calculated by subtracting the Baseline Metadata Length from Hop_ML.
```
mickeyspiegel: I think the text about Domain Specific ID that is not known to the node should go after the text about Domain Specific ID that is known to the node.

More importantly, I want to make it clear that when the Domain Specific ID is not known, as an alternative to padding with all 1s, the node could not add any metadata at all. That is why I suggested the following text:

If the Domain Specific ID does not match any Domain ID known to this node, then
the node is required to either:

- Pad the node's INT Metadata stack with the special all-ones reserved value for a Domain Specific Metadata length, calculated by subtracting from the Hop ML a length computed from all bits in the 16-bit INT Instruction field, or

- Skip INT processing altogether and not insert any metadata into the packet

mickeyspiegel (changes requested):

telemetry/specs/INT.mdk
rsivakolundu: Makes sense. Done.

telemetry/specs/INT.mdk
rsivakolundu: Good Point. Fixed it.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
```diff
@@ -435,6 +440,13 @@ is shared between multiple queues. The format of this 4-octet metadata field is
 implementation specific and the metadata semantics YANG model shall describe the format 
 and units of this metadata field in the metadata stack.
 
+* Buffer occupancy
+  - The build-up of traffic in the buffer (in bytes, cells, or packets) that the
+INT packet observes in the device while being forwarded. Use case is when the buffer 
+is shared between multiple queues. The format of this 4-octet metadata field is switch 
+implementation specific and the metadata semantics YANG model shall describe the format 
+and units of this metadata field in the metadata stack.
+
```
mickeyspiegel: The change above is still a duplicate and should be deleted.

telemetry/specs/INT.mdk
```diff
@@ -306,25 +309,27 @@ at the monitoring system to collate reports from multiple devices.
...
+as all traffic. 
+
 
-The XD mode doesn't make sense for synthetic traffic. The MX mode is
-technically possible but doesn't add much value. Hence we define two MD modes
-as follows.
+The only difference between live traffic and Synthetic traffic is that 
+INT-Sink nodes may need to discard synthetic traffic after extracting 
+the collected INT data as opposed to forwading the traffic. This is 
+indicated by using the 'C' bit of the INT-Header to mark relavant packets
```
mickeyspiegel: typo: s/relavant/relevant

telemetry/specs/INT.mdk
```diff
@@ -306,25 +309,27 @@ at the monitoring system to collate reports from multiple devices.
...
 
-* INT-CLONE-MD (eMbed Data): INT Source devices create extra traffic through
-cloning the original packets and then embedding the INT instructions and metadata into
-the clones only. INT Transit devices accumulate metadata only on the clones,
-INT Sink devices may send telemetry reports before discarding the clones.
-Compared to INT-MD, INT-CLONE-MD produces extra traffic
-and may not capture actual network state that the orignal packets experience, while
-this mode avoids direct modification on the original data packets.
-This mode is similar to IFA [^IFA].
+All INT modes may be used on these Syntehtic/probe packets, as decided by the INT-Source node.
```
mickeyspiegel: typo: s/Syntehtic/Synthetic

telemetry/specs/INT.mdk
```diff
@@ -306,25 +309,27 @@ at the monitoring system to collate reports from multiple devices.
...
-and may not capture actual network state that the orignal packets experience, while
-this mode avoids direct modification on the original data packets.
-This mode is similar to IFA [^IFA].
+All INT modes may be used on these Syntehtic/probe packets, as decided by the INT-Source node. 
+Specifically the **INT-MD** (eMbed Data) mode applied to Synthetic or probe packets allows 
+functionality similar to **IFA**[^IFA] 
 
-* INT-PROBE-MD (eMbed Data): similar to INT-CLONE-MD, but INT is applied to
-special probe packets instead of clone packets.
+It is likely that sytnetic traffic created by cloning would be discraded at the Sink,
```
mickeyspiegel: typos:
s/sytnetic/synthetic
s/discraded/discarded

telemetry/specs/INT.mdk
```diff
@@ -306,25 +309,27 @@ at the monitoring system to collate reports from multiple devices.
...
+while Probe packets might be marked for forwatrding or discarding, depending on the
```
mickeyspiegel: typo: s/forwatrding/forwarding

telemetry/specs/INT.mdk
```diff
@@ -306,25 +309,27 @@ at the monitoring system to collate reports from multiple devices.
...
+use-case. It is the responsibility of the INT-Source node to mark packets correctly 
+to determine if the INT-Sink will forward or discard packets after extracting the 
+INT Data colected along the path.
```
mickeyspiegel: typo: s/colected/collected

telemetry/specs/INT.mdk
```diff
@@ -1088,18 +1104,21 @@ header.
     metadata header.
 * Summary of the field usage
   - The INT Source must set the following fields:
-    - Ver, Rep, C, M, Per-hop Metadata Length, Remaining Hop Count,
+    - Ver, C, M, Per-hop Metadata Length, Remaining Hop Count,
       and Instruction Bitmap.
     - INT Source must set all reserved bits to zero.
+    - INt Source may set the Domain-specific fileds.
```
mickeyspiegel: typos:
s/INt/INT
s/fileds/fields

telemetry/specs/INT.mdk
mickeyspiegel: This change adds a fourth 0 before the Reserved field. That does not belong.

telemetry/specs/INT.mdk
```diff
@@ -1194,9 +1212,11 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
     0                   1                   2                   3
```
mickeyspiegel, line 1204: The length above in the shim header does not seem correct.

telemetry/specs/INT.mdk
```diff
@@ -948,38 +963,16 @@ INT Metadata Header and Metadata Stack:
...
-  - C (1b): Copy.
-    - If replication is requested for data packets, the INT Sink must be
-able to distinguish the original packet from replicas so that it can forward
-only original packets up the protocol stack, and drop all the replicas. The C
-bit must be set to 1 on each copy, whenever an INT hop replicates a packet.
-The original packet must have C bit set to 0.
-    - C bit must be set to 0 in the original packet by INT source
+  - Ver (4b): INT metadata header version. Should be 2 for this version.
+  - Res (2b): Reserved
+  - C   (1b): Copy/Clone. INT Sink should Discard the packet after Extracting INT data
```
mickeyspiegel: I wonder about calling this "Copy/Clone". Is a probe packet a copy or a clone?
In the IETF I suggested calling it "active".
If you have any better suggestions, I am open to hearing them.

mickeyspiegel (changes requested):
I see two major differences from the IETF's IOAM namespace beyond the additional fields:
1. When a transit node receives an INT packet with a Domain Specific ID that it does not recognize, is it still allowed to add its metadata to the packet?
2. Does the Domain Specific ID have any bearing on the semantics and/or metadata units? In the IETF this seems to be the main point of namespace.
It would be good to hash this out further, either aligning with the IETF or clarifying at least to ourselves if not in the spec, why should this be different?

Remaining changes are minor.
Remove some duplicate text.
Fix typos.
Fix packet format changes that no longer apply.
Check packet length in example.

telemetry/specs/INT.mdk
jklr: how about D for Discard?
Sink should discard the pkt (after generating a report) no matter what the reason is.

mickeyspiegel: Works for me

jklr:
@rsivakolundu 
I wonder if we can prioritize this PR and merge first than others. 
The change of this PR is applied to every INT mode (old and new) and the examples. 
If the key section is updated and merged into master, I can update the other examples (under different transports) and add new examples accordingly.

telemetry/specs/INT.mdk
rsivakolundu: Works for me as well. Changed it to "D" in the doc.

telemetry/specs/INT.mdk
rsivakolundu: Fixed.

telemetry/specs/INT.mdk
rsivakolundu: Corrected the typo.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Fixed the typo.

telemetry/specs/INT.mdk
rsivakolundu: Corrected the typos.

telemetry/specs/INT.mdk
rsivakolundu: Corrected the typo.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Deleted the duplicate paragraph.

telemetry/specs/INT.mdk
rsivakolundu: Removed the new flag bit. We need to discuss the need for a bit to indicate the VxLAN vs VxLAN GPE encap of the incoming packet. Do we need that indication in the header for the sink node processing.

jklr: We're addressing the GPE discussion and flag in #58. 
Let's merge this PR first into mater and rebase #58.

jklr:
Thanks @rsivakolundu for the fixes. They look good to me.
Can you pls comment on @mickeyspiegel 's questions below?
I put my answers.

> I see two major differences from the IETF's IOAM namespace beyond the additional fields:
> 
> 1. When a transit node receives an INT packet with a Domain Specific ID that it does not recognize, is it still allowed to add its metadata to the packet?

[JK] I didn't think through but the answer depends on the Yang-model-based query mechanism. 
Yes, if the collector can ping each device and get the metadata semantics.
No, if not. 
Agree we better clarify this in the spec.

> 2. Does the Domain Specific ID have any bearing on the semantics and/or metadata units? In the IETF this seems to be the main point of namespace.
>    It would be good to hash this out further, either aligning with the IETF or clarifying at least to ourselves if not in the spec, why should this be different?

[JK] associating the domain ID to the metadata semantics seems a good idea to me.

mickeyspiegel (approved):

jklr:
Let's merge it and address the domain ID questions in a separate PR on top of master.

PR #55, opened 2018-10-02
Buffer Occupancy

rsivakolundu:
Added instruction bit for buffer occupancy


telemetry/specs/INT.mdk
```diff
@@ -910,6 +914,170 @@ from (shim header length \* 4).
...
+`
+
+* INT metadata header is 8 bytes long followed by a stack of INT metadata.
+  Each metadata is either 4 bytes or 8 bytes in length. Each INT hop adds
+  the same length of metadata. The total length of the metadata stack is
+  variable as different packets may traverse different paths and hence
+  different number of INT hops.
+
+* The fields in the INT metadata header are interpreted the following way:
+  - Ver (4b): INT metadata header version. Should be 1 for this version.
```
mhira1: Version must be 2 in this paragraph

telemetry/specs/INT.mdk
```diff
@@ -318,6 +318,10 @@ simply leaves those decisions to device vendors.
   - The build-up of traffic in the queue (in bytes, cells, or packets) that the
 INT packet observes in the device while being forwarded.
 
+* Buffer occupancy
+  - The build-up of traffic in the shared buffer (in bytes, cells, or packets) that the
+INT packet observes in the device while being forwarded.
+
```
mhira1, line 335: Perhaps remove the word "shared". Reporting both queue occupancy and buffer occupancy makes sense only if the buffer is shared, but technically the buffer could be a per-queue buffer, just that in this case, both values reported will be the same. INT source does not know whether downstream switches have shared buffers or not. It can set both bits. Switches that have per-queue buffers (unlikely, switches typically use shared buffers) will just report the same value for both instructions.

mhira1: May be word this differently, to say this instruction is to get buffer occupancy. Use-case is when buffer is shared.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +915,170 @@ from (shim header length \* 4).
...
+      inserted at one or more hops and corrective action such as reconfiguring
+      MTU at some links may be needed, particularly when INT switches are not
+      participating in path MTU discovery. The M bit is not aimed at readily
+      identifying which switch(es) did not insert INT metadata due to egress MTU
+      limitation. In theory, if this does not occur at consecutive hops,
+      it may be possible for the monitoring system to derive which
+      switch(es) set the M bit based on knowledge of the network topology
+      and "Switch ID, Ingress port ID, Egress port ID" tuples in the INT
+      metadata stack.
+  - R (9b): Reserved bits.
```
mhira1: Number of reserved bits is still 10 correct? I think this text with 9 bits is coming from your other patch in which you are using one of the reserved bits for "Sink" bit. There is no change in number of reserved bits between v1 and v2 in this patch.

mhira1 (commented):
Minor change needed.

telemetry/specs/INT.mdk
```diff
@@ -910,6 +915,170 @@ from (shim header length \* 4).
 For INT over Geneve it is 8 bytes subtracted from (length in Geneve tunnel
 option header \* 4).
 
+
+INT Metadata Header and Metadata Stack (Version = 2):
```
jklr: @mhira1 I was meaning to propose make a 1.x version cut before merging these changes, and move INT.mdk to version2 draft. Then we don't need to list both v1 and v2 in the same doc. How do you think?

mhira1: Makes sense. Let's remove this second paragraph on v2 header then.

rsivakolundu:
Ok. I will add buffer occupancy to ver 1 and remove ver 2 details.

-Ramesh

jklr:
@rsivakolundu   hold on, pls. This buffer occupancy change is NOT backward compatible, right? Hence it should be added only to v2. Let's discuss version maintenance policy tomorrow and then make further changes on the PRs. I don't want you to waste time during PTO.

telemetry/specs/INT.mdk
rsivakolundu: Removed the reference to shared.

mickeyspiegel: Minor editorial rewording:
The use case is when the buffer is shared between multiple queues.

telemetry/specs/INT.mdk
```diff
@@ -1279,6 +1289,7 @@ struct headers {
     int_egress_tstamp_t         int_egress_tstamp;
     int_level2_port_ids_t       int_level2_port_ids;
     int_egress_port_tx_util_t   int_egress_port_tx_util;
+    int_b_occupancy_t		int_b_occupancy;
```
mickeyspiegel, line 1313: nit: fix indent

telemetry/specs/INT.mdk
```diff
@@ -1610,6 +1621,11 @@ control EgressDeparserImpl(packet_out packet,
             ck.add({hdr.int_egress_port_tx_util.egress_port_tx_util});
         }
 
+ 	if (hdr.int_b_occupancy.isValid()) {
```
mickeyspiegel: nit: fix indent

telemetry/specs/INT.mdk
```diff
@@ -1656,6 +1672,7 @@ control EgressDeparserImpl(packet_out packet,
         packet.emit(hdr.int_egress_tstamp);
         packet.emit(hdr.int_level2_port_ids);
         packet.emit(hdr.int_egress_port_tx_util);
+	packet.emit(hdr.int_b_occupancy);
```
mickeyspiegel: nit: fix indent

telemetry/specs/INT.mdk
rsivakolundu: Added the minor editorial change suggested.

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
```diff
@@ -2020,6 +2037,21 @@ INT packet was served.  The calculation mechanism of this value is device
 specific.
 * Queue drop count
   - Total # of packets dropped from the queue
+  
+* Buffer occupancy
+  - The build-up of traffic in the buffer (in bytes, cells, or packets) that the
+INT packet observes in the device while being forwarded. Use case is when buffer is
+shared between multiple queue.
```
mhira1, line 2066: typo: multiple queues

telemetry/specs/INT.mdk
```diff
@@ -2020,6 +2037,21 @@ INT packet was served.  The calculation mechanism of this value is device
...
+* Instantaneous buffer occupance
```
mhira1: typo: occupance -> occupancy

telemetry/specs/INT.mdk
```diff
@@ -2020,6 +2037,21 @@ INT packet was served.  The calculation mechanism of this value is device
...
+  - The instantaneous value (in bytes, or cells) of the buffer occupancy the INT
+packet has observed in the device while being forwarded.  The units used need
+not be consistent across an INT domain, but care must be taken to ensure that
+there is a known, consistent mapping of {device, buffer} values to their
+respective unit {bytes, cells}.
+* Average buffer occupancy
+  - The average value (in bytes or cells) of the buffer occupancy 
+the INT packet was observed. The calculation mechanism of this value is device
```
mhira1, line 2078: grammar: average value of buffer occupancy that the INT packet observed.

mhira1 (commented):
Ramesh, I looked through the edits again, marked some typos. Also, at the previous meeting, we decided to include both buffer ID and buffer occupancy into the 32-bit space. Please make that change as well.

telemetry/specs/INT.mdk
```diff
@@ -316,7 +316,16 @@ simply leaves those decisions to device vendors.
 
 * Queue occupancy
   - The build-up of traffic in the queue (in bytes, cells, or packets) that the
-INT packet observes in the device while being forwarded.
+INT packet observes in the device while being forwarded. The format of this 4-octet 
+metadata field is switch implementation specific and the YANG model shall describe
```
jklr: Can we add a reference to the YANG model, since this is the first time mentioning it?
We can reword 'the YANG model' to 'the metadata semantics YANG model' and put a reference to :
https://github.com/p4lang/p4-applications/blob/master/telemetry/code/models/p4-dtel-metadata-semantics.yang

telemetry/specs/INT.mdk
```diff
@@ -752,12 +761,12 @@ hop-by-hop INT header must fit in a single Geneve option.
 In this section, we define the format for INT hop-by-hop metadata headers,
 and the metadata itself.
 
-INT Metadata Header and Metadata Stack:
+INT Metadata Header and Metadata Stack (Version = 1):
```
jklr: Shall we remove "(Version = 1)" from above? We are not going to present two versions in one document. 
The version number in the header format below is good to have though.

telemetry/specs/INT.mdk
```diff
@@ -854,6 +863,7 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
   - bit5: Egress timestamp
   - bit6: Level 2 Ingress Port ID + Egress Port ID (4 bytes each)
   - bit7: Egress port Tx utilization
+  - bit8: Buffer ID (8 bits) + Buffer Occupancy (24 bits)
```
jklr: Shall we clarify that the 8b:24b layout is one example and switch implementation specific? This is the same for bit3 (queue occupancy field).

telemetry/specs/INT.mdk
```diff
@@ -1219,6 +1229,10 @@ header int_egress_port_tx_util_t {
     bit<32> egress_port_tx_util;
 }
 
+header int_b_occupancy_t {
+    bit<32> q_occupancy;
```
jklr: Please make it consistent with the (example) layout provided in line 866, which has 8b buffer id.

telemetry/specs/INT.mdk
```diff
@@ -1610,6 +1625,11 @@ control EgressDeparserImpl(packet_out packet,
             ck.add({hdr.int_egress_port_tx_util.egress_port_tx_util});
         }
 
+        if (hdr.int_b_occupancy.isValid()) {
+            ck.add({
+                    hdr.int_b_occupancy.b_occupancy
```
jklr, line 1648: Same here. If we want buffer_id in line 866, let's make the code consistent.

telemetry/specs/INT.mdk
```diff
@@ -2021,6 +2042,23 @@ specific.
 * Queue drop count
   - Total # of packets dropped from the queue
 
+
+The buffer occupancy INT packet observes in the device while being forwarded. Use case is when buffer is
```
jklr: Let's start the sentence with sth like "The metadata below are introduced to capture the buffer occupancy INT packet ... ".

jklr (approved):
@rsivakolundu thanks for the PR. I suggested several editorial changes. Please check the last two nit comments from Mukesh.

telemetry/specs/INT.mdk
rsivakolundu: Fixed the typo.

telemetry/specs/INT.mdk
rsivakolundu: Added that

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Shall we discuss this in Thursday meeting.

telemetry/specs/INT.mdk
rsivakolundu: Let us discuss the semantics of Queue and Buffer occupancy. I will make it consistent after the discussion.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Fixed

telemetry/specs/INT.mdk
rsivakolundu: Fixed

telemetry/specs/INT.mdk
rsivakolundu: Removed (Version = 1)

telemetry/specs/INT.mdk
rsivakolundu: Added Buffer ID (8 bits) and Buffer Occupancy (24 bits)

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
rsivakolundu: Added text to support other semantics for Queue Occupancy and Buffer Occupancy.

jklr:
@rsivakolundu your latest commit merged domain specific extension to this branch/PR!

rsivakolundu:
@JK - I don't see it in my branch that way.

-Ramesh

jklr:
@rsivakolundu hmm I think there is a github bug. The bottom most commit in this PR's commit tab showed domain specific changes 2 hours ago. I double checked at that time.
 It's not like that any more.

jk:
@rsivakolundu @jklr != @jk

PR #57, opened 2018-10-18
Add 16-bit Sink Indication field to INT header

mickeyspiegel:
A new 16-bit Sink Indication field is proposed to be added to the INT header, placed in the previous 16-bit Reserved field following the Instruction Bitmap in the INT header. This field allows the source and transit switches to notify the sink switch about the collected network state in a concise manner.

While INT allows network state to be collected on a per packet basis, for many packets the network state may not differ significantly from the state collected in previous packets belonging to the same flow. The sink may choose to report only a subset of packets, or it may choose to aggregate the collected network state.

The Sink Indication lets the sink switch look at only one field when deciding how to process the telemetry information from the packet, without having to scrape over each field added at each hop, just to decide whether and how to process the collected network state.

The Sink Indication is initially set by the source switch, and is updated by each transit switch, taking into account the received value as well as the local state added by that transit switch.


rsivakolundu:
Are we changing this in Ver=2 or Ver=1. Also, wasn't this field to be called Domain Specific ID. I am confused at this stage.

jklr:
We will do it in Ver2, by using domain specific extension. I think we can just close this PR. Mickey can decide.

PR #58, opened 2019-01-17
Add bit indicating whether orig packet used VXLAN or VXLAN-GPE encap

mickeyspiegel:
Include guidelines suggesting that the sink should ignore this bit in
mixed v1.0 / v1.1 environments.


telemetry/specs/INT.mdk
```diff
@@ -711,6 +711,26 @@ INT shim header for VXLAN GPE encapsulation:
 * Type: This field indicates the type of INT Header following the shim header.
 Two Type values are used: one for the hop-by-hop header type and the other for
 the destination header type (See Section [#sec-int-header-types]).
+* G: Indicates whether the original packet (before insertion of INT headers
+and metadata) used a VXLAN or VXLAN GPE encapsulation.
+  - 0: Original packet used VXLAN encapsulation.
+  - 1: Original packet used VXLAN GPE encapsulation.
```
mhira1: Wondering whether we should flip this and use 0 to indicate the original packet was VXLAN GPE, i.e. no conversion is required and 1 to indicate that original packet was VXLAN, i.e. yes conversion is required at INT sink. Thus INT sink does VXLAN GPE to VXLAN conversion if this bit is set to 1.

mhira1 (commented):
Looks good, minor comment inline.

jklr:
@mickeyspiegel
This is a reminder of the AR from the May meeting: All text around 1.0/1.x interoperability needs to be modified as this is now going into version 2.0. Mickey to revise, Mukesh to review and merge.

telemetry/specs/INT.mdk
```diff
@@ -711,6 +711,26 @@ INT shim header for VXLAN GPE encapsulation:
...
+    value of this bit and rely on local configuration to determine whether to
+    convert the VXLAN GPE encapsulation back to a VXLAN encapsulation.
+    Typically in such a mixed or v1.0 environment, INT would only be inserted
+    into one type of encapsulation but not the other, e.g. VXLAN but not
+    VXLAN GPE since the INT sink always converts VXLAN GPE packets with INT
+    back to the VXLAN encapsulation.
+  - In an all v1.1 environment, both packets received with VXLAN encapsulation
+    and VXLAN GPE encapsulation could have INT inserted, with the INT sink
+    relying on the value of this bit to determine whether to carry out the
+    conversion from VXLAN GPE back to VXLAN.
```
mickeyspiegel: Since this is now aimed at the INT v2.0 spec, the two bullet points about v1.0, v1.1 and interop can just be removed.

jklr: +1

telemetry/specs/INT.mdk
jklr: I don't have a very strong preference but using 1 to indicate the need for conversion seems a bit more indicative and useful for debugging. Again, no strong opinion.

telemetry/specs/INT.mdk

mickeyspiegel:
Open Issues:

1. Should we reduce the INT-Type field to 4 bits as JK recently proposed?
2. Should we move the G bit to the Type byte, or leave it in the Reserved byte?

Note that the VXLAN GPE shim header format in draft-ietf-nvo3-vxlan-gpe-09 defines:
```
   |      Type     |    Length     |    Reserved   | Next Protocol |
```
where
Type:  This field MAY be used to identify different messages of this protocol.
Reserved:  The use of this field is reserved to the protocol defined in this message.

jklr:
@mickeyspiegel as for the shim hdr format, I think the consensus from the last meeting is to follow the GPE convention, deviating from the INT convention.

jklr (approved):

PR #61, opened 2019-10-22
Telemetry report v2.0

rsivakolundu:
Added changes for v2.0 report.


telemetry/specs/INT.mdk
```diff
@@ -879,7 +879,7 @@ hop-by-hop INT header must fit in a single Geneve option.
 In this section, we define the format for INT hop-by-hop metadata headers,
 and the metadata itself.
 
-INT Metadata Header and Metadata Stack:
+INT Metadata Header and Metadata Stack (Version = 1):
```
jklr: I think we decided to keep just v2 contents, instead of keeping both v1 and v2 in the same version of spec. We will have all previous versions of the specs in the github and p4.org spec link.

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,13 +440,13 @@ The Destination Port is user configurable. The expectation is that the same
 Destination Port value will be used for all telemetry reports in a
 particular deployment.
 
-## Telemetry Report Header (16+ octets)
+## Telemetry Report Header (Ver 1.0) (16+ octets)
```
jklr: same here. we can remove v1.0 contents.

telemetry/specs/telemetry_report.mdk
```diff
@@ -316,6 +314,24 @@ Telemetry Report spec.
 ~
 [individual-report-arch]: images/telemetry_arch_individual.jpg { width: 6.0in }
 
+In the INT-MX (eMbed instruct(X)ions) modes, as defined in the INT spec,
```
mickeyspiegel: nit: s/modes/mode

telemetry/specs/telemetry_report.mdk
```diff
@@ -316,6 +314,24 @@ Telemetry Report spec.
...
+the source node embeds instructions in the INT headers. Upon receipt of a packet 
+with this INT type, each network device in the flow path generates its own own 
+telemetry reports, as shown in Figure [#fig-INT-MX-arch]. The distributed telemetry 
+monitoring system will receive reports from different network devices, each describing 
+the telemetry metadata (such as switch IDs, port IDs, latency)for one hop. There is no
```
mickeyspiegel: nit: s / latency)for / latency) for

telemetry/specs/telemetry_report.mdk
```diff
@@ -316,6 +314,24 @@ Telemetry Report spec.
...
 [individual-report-arch]: images/telemetry_arch_individual.jpg { width: 6.0in }
 
+In the INT-MX (eMbed instruct(X)ions) modes, as defined in the INT spec,
+the source node embeds instructions in the INT headers. Upon receipt of a packet 
+with this INT type, each network device in the flow path generates its own own 
+telemetry reports, as shown in Figure [#fig-INT-MX-arch]. The distributed telemetry 
+monitoring system will receive reports from different network devices, each describing 
+the telemetry metadata (such as switch IDs, port IDs, latency)for one hop. There is no 
+change to data packets, besides the source node embedding the INT Header with
+instrucing, traversing the network. The sink node removes the INT Header from
```
mickeyspiegel: s / instrucing, traversing the network / instructions

Either delete "traversing the network", or put it before the first ","

telemetry/specs/telemetry_report.mdk
```diff
@@ -337,11 +353,11 @@ of the aggregated telemetry metadata from all hops across the network.
 [int-md-arch]: images/telemetry_arch_figure.jpg { width: 7.5in }
 
 In order to reduce complexity at the sink switch, some telemetry reports
-may include embedded telemetry metadata intermingled with the original
-packet headers. This simplifies generation of telemetry reports due to
-receipt of data packets with embedded telemetry metadata. The telemetry
-data plane specification such as INT or IOAM specifies the format for
-this portion of the telemetry metadata. This approach reduces data plane
+may include embedded telemetry metadata intermingled with the details of
+orignial packet headers. This simplifies generation of telemetry reports
```
mickeyspiegel: typo: s/orignial/original

telemetry/specs/telemetry_report.mdk
```diff
@@ -316,6 +314,24 @@ Telemetry Report spec.
...
+monitoring system will receive reports from different network devices, each describing 
+the telemetry metadata (such as switch IDs, port IDs, latency)for one hop. There is no 
+change to data packets, besides the source node embedding the INT Header with
+instrucing, traversing the network. The sink node removes the INT Header from
+such packets. When using INT-MX mode, the telemetry metadata precedes the 
+details of the original packet headers within the telemetry report.
+
+
+~ Figure { #fig-INT-MX-arch; caption: "INT-MX - Telemetry Architecture with reports \
+   generated by all switches." }
```
mickeyspiegel, line 329: The caption text should be aligned for all three modes. Also INT-MX needs to be removed from the caption for the previous figure. Due to this change, it only covers INT-XD.

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+* Length
+  : Indicates the length of the telemetry report header in multiples of
+    4 octets, including the variable optional metadata.
+
+* NProt: Next Protocol
+  :
+    * 0: Ethernet
+    * 1: IPv4
+    * 2: IPv6
+    * 3: Telemetrty Extensions
```
mickeyspiegel: typo: s/Telemetrty/Telemetry

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+    packet matched a flow watchlist somewhere (in case of INT or IOAM) or
+    locally (in case of postcard). The report might include INT or IOAM
+    metadata beyond the inner ethernet header. Other telemetry reports are
+    likely to be received for the same tracked flow, from the same network
+    device and (in case of drop reports, postcard or path changes) from
+    other network devices.
+    
+  * L: Last Telemetry Record
+    : Indicates that this telemetry report header is the last record of this telemetry
+      report.
```
mickeyspiegel: What does it mean to have multiple records of a telemetry report?
What is the common part that links them together?

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+    (Switch id, hw_id) to a particular telemetry report destination. This
+    can be used to detect loss of telemetry reports before they reach their
+    intended destination.
+
+* Ingress Timestamp
+  : The device local time when the packet was first received on the ingress
+    physical or logical port, in nanoseconds.
+    
+* Variable Optional Metadata
+  : The metadata corresponding to the RepMdBits, 4 octets for each bit.
```
mickeyspiegel: Except 8 bits for bit 7

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+  
+* Extension Length
+  : Indicates the length of the domain specific metadata included in the 
+    this record of the telemetry report.
```
mickeyspiegel: In multiples of 4 octets?

We should align the terminology. If we go with "domain specific metadata", then we should change "Extension Data" below to "Domain Specific Metadata".

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+* Domain Specific ID
+  : The unique ID of the INT Domain
+
+* Domain Specific Md Bits
+  : Bitmap that indicates which optional domain specific metadata is 
+    present in the record of the telemetry report. Each bit represents 
+    4 octets of domain specific optional metadata.
+    
+* Domain Specific Md Status
+  : Bitmap that indicates the domain specific metadata status.
```
mickeyspiegel, line 572: What is a domain specific metadata status?

mickeyspiegel (changes requested):
A few more things that need to be taken care of:

- Update references to IOAM drafts, INT spec, SAI DTel to the latest. Add a reference to SAI TAM.
- The scope section should refer to INT-MD vs INT-MX, INT-XD
- In many cases "INT" should be replaced by "INT-MD"
- Rather than describing separate flow and drop watchlists, it should refer to a telemetry watchlist that can specify actions to monitor flows, monitor drops, or monitor both flow and drops.
- Clear the page break between "Dropped Packets" and "Congested Queues"

telemetry/specs/telemetry_report.mdk
```diff
@@ -510,6 +526,132 @@ particular deployment.
...
+    * 1: IPv4
+    * 2: IPv6
+    * 3: Telemetrty Extensions
+    * 4-14: Reserved for future use
+    * 15: None
+
+* RepMdBits: Report Metadata Bits
+  : Bitmap that indicates which optional metadata is present in the
+    telemetry report header. Each bit represents 4 octets of optional
+    metadata.
```
mickeyspiegel: except for bit 7 which represents 8 octets of optional metadata.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Removed

telemetry/specs/telemetry_report.mdk
rsivakolundu: Corrected.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Corrected.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Corrected.

telemetry/specs/telemetry_report.mdk
rsivakolundu: I think I fixed it. Review and comment please.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Corrected.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Removed the "L" bit. Multiple records in a telemetry report can be done differently. We can discuss it at the next WG meeting.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Changed everything to be Domain Specific Metadata.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Added

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
```diff
@@ -1035,6 +1035,7 @@ INT-MD header must fit in a single Geneve option.
 In this section, we define the format for INT-MD metadata headers,
 and the metadata itself.
 
+
```
mickeyspiegel: Please revert this whitespace change in INT

telemetry/specs/telemetry_report.mdk
```diff
@@ -1,8 +1,8 @@
-Title : Telemetry Report Format Specification
+Title : Telemetry Report Format Specification Version 2.0
 Title Note : Working draft. Note: consider using tagged versions for implementation.
 Title Footer : 2018-04-20
 Author : The P4.org Applications Working Group
-Affiliation: Contributions from *Barefoot Networks, VMware, Xilinx*
+Affiliation: Contributions from *Barefoot Networks, Cisco Systems, VMware, Xilinx*
```
mickeyspiegel: "Barefoot Networks/Intel"

telemetry/specs/telemetry_report.mdk
```diff
@@ -1,8 +1,8 @@
-Title : Telemetry Report Format Specification
+Title : Telemetry Report Format Specification Version 2.0
 Title Note : Working draft. Note: consider using tagged versions for implementation.
```
mickeyspiegel, line 2: This "Title Note" content will have to go when we finalize. I think "Version 2.0" belongs here rather than in the title itself.

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+* Sequence Number (22b): Sequence Number
+  : Reflects the sequence of reports from a specific combination of
+    (Node id, hw_id) to a particular telemetry report destination. This
+    can be used to detect loss of telemetry reports before they reach their
+    intended destination.
 
-* Ver: Version
-  : This specification defines **version 1**.
+* Node ID (32b): Node ID
+  : The unique ID of a switch (generally administratively assigned). Node IDs
```
mickeyspiegel: We generalized to "node" so that it is not just switches.
Network devices?
Mention switch and NIC as possible examples?

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+    must be unique within a management domain.
 
-* Length
-  : Indicates the length of the telemetry report header in multiples of
-    4 octets, including the variable optional metadata.
+* RepType (4b): Report Type
+  : Type of Telemetry report
+    * 0: INT v2.0
```
mickeyspiegel: v2.0 or 2.1

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+    * 2 – 15: Reserved
+
+* InType (4b): Inner Type
+  : Type of data embedded after Variable Optional Baseline & DS Metadata.
+    * 0: None
+    * 1: Domain Specific Extension Data
+    * 2: Ethernet
+    * 3: IPv4
+    * 4: IPv6
+    * 5-7: Reserved.
```
mickeyspiegel: 5-15: Reserved

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
-    * 0: Ethernet
-    * 1: IPv4
-    * 2: IPv6
+* Q (1b): Congested Queue Association
+  : Indicates the presence of congestion on a monitored queue.
 
-* RepMdBits: Report Metadata Bits
+* F (1b): Tracked Flow Association
+  : Indicates that this telemetry report is for a tracked flow, i.e. the
+    packet matched a flow watchlist somewhere (in case of INT or IOAM) or
```
mickeyspiegel: INT-MD, INT-MX, or IOAM

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+    locally (in case of postcard). The report might include INT or IOAM
```
mickeyspiegel: s/postcard/INT-MX
s/INT/INT-MD

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+    metadata beyond the inner ethernet header. Other telemetry reports are
+    likely to be received for the same tracked flow, from the same network
+    device and (in case of drop reports, postcard or path changes) from
```
mickeyspiegel: s/postcard/INT-MX, INT_XD

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
   : Bitmap that indicates which optional metadata is present in the
     telemetry report header. Each bit represents 4 octets of optional
-    metadata.
-    * bit 0 (MSB): Ingress port id (16 bits) + Egress port id (16 bits)
-    * bit 1: Hop latency
-    * bit 2: Queue ID (8 bits) + Queue occupancy (24 bits)
-    * bit 3: Egress Timestamp
-    * bit 4: Queue ID (8 bits) + Drop reason (8 bits) + Padding (16 bits)
-    * bit 5: Egress port tx utilization
+    metadata, except for bit 4,5 & 6 which represents 8 octets of optional metadata.
```
mickeyspiegel: nit: s/bit 4,5/bits 4, 5

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+    * bit 0 (MSB): Reserved
+    * bit 1: Level 1 Ingress Interface ID (16 bits) & Egress Interface ID (16 bits)
+    * bit 2: Hop Latency
+    * bit 3: Queue ID (8 bits) + Queue Occupancy (24 bits)
+    * bit 4: Ingress Timestamp (64 bits)
+    * bit 5: Egress Timestamp (64 bits)
+    * bit 6: Level 2 Ingress Interface ID (32 bits) + Egress Interface ID (32 bits)
+    * bit 7: Egress Port TX Utilization
+    * bit 8: Buffer ID (8 bits) + Buffer Occupancy (24 bits)
+    * bit 9-14: Reserved. Should be set to zero.
```
mickeyspiegel: upon transmission and ignored upon reception

telemetry/specs/telemetry_report.mdk
```diff
@@ -473,48 +548,38 @@ particular deployment.
...
 
     See the INT specification [^INT] for definitions of the remaining metadata.
+    
+* Domain Specific ID (16b)
+  : The unique ID of the INT Domain
+
+* DSMdbit (16b): Domain Specific Md Bits
+  : Bitmap that indicates which optional domain specific metadata is 
+    present in the record of the telemetry report. Each bit represents 
+    4 octets of domain specific optional metadata.
```
mickeyspiegel: Did you want to nail this down to 4 octets each?
Or leave it flexible and rely on MD Length to pad if necessary?

mickeyspiegel: I don't know where to specify this since it falls in between INT Data Plane and Telemetry Report. I guess this is as good a place as any.

When using INT-MD or INT-MX, if the Domain Specific ID does not match any Domain ID known to this node, then the node may either:

- Copy the INT Hop ML to the Telemetry Report MD Length, copy the INT DS Instruction to the Telemetry Report DSMdbit field, and pad the Variable Optional Baseline & DS Metadata with the special all-ones reserved value out to MD Length, or
- Set the Telemetry Report DSMdbit field to zero and rederive the Telemetry Report MD Length from RepMdBits.

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+      |  Ver  |   hw_id   |           Sequence Number                 |
       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-      |                  Variable Optional Metadata                   |
-      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-`
+      |                            Node id                            |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+      |RepType| InType| Report Length |   MD Length   |D|Q|F|I|  Rsvd |  |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  R
+      |            RepMdbit           |      Domain Specific ID       |  E
```
mickeyspiegel: nit: We should be consistent about singular versus plural for bit fields, referring to RepMdBit and DSMdbit here, and the definitions below. I prefer plural but I don't feel strongly about it.

telemetry/specs/telemetry_report.mdk
```diff
@@ -473,48 +548,38 @@ particular deployment.
...
-* Ingress Timestamp
-  : The device local time when the packet was first received on the ingress
-    physical or logical port, in nanoseconds.
 
 ## Flow Identification
 
 There is no explicit metadata defined for flow identification. The
 expectation is that packet headers will be included in the telemetry
-report, allowing the distributed telemetry monitoring system to categorize
+report or domain specific flow identification metadata will be included,
```
mickeyspiegel: nit: This is a little hard to parse. How about:

The expectation is that either:

- a truncated packet fragment including packet headers will be included in the telemetry report,
  allowing the distributed telemetry monitoring system to categorize and identify flows in any
  manner that it desires, or

- domain specific flow identification metadata will be included.

mickeyspiegel (changes requested):
I think we are close.

My high level comments from Jan 14 are all still applicable. Let me know if you want me to handle them and commit them to this pull request.

telemetry/specs/INT.mdk
rsivakolundu: Deleted the white space.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
rsivakolundu: v2.0 or later. Will that work?

telemetry/specs/telemetry_report.mdk
rsivakolundu: Fixed

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
rsivakolundu: Added

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Yes

telemetry/specs/telemetry_report.mdk
rsivakolundu: Added

telemetry/specs/telemetry_report.mdk
rsivakolundu: Agree. Fixed it to be RepMdBits and DSMdBits consistently everywhere.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Agree. Modified the text.

rsivakolundu:
Addressed all the changes requested.

telemetry/specs/telemetry_report.mdk
```diff
@@ -511,11 +511,11 @@ particular deployment.
 
 * F (1b): Tracked Flow Association
   : Indicates that this telemetry report is for a tracked flow, i.e. the
-    packet matched a flow watchlist somewhere (in case of INT or IOAM) or
-    locally (in case of postcard). The report might include INT or IOAM
+    packet matched a flow watchlist somewhere (in case of INT-MD, INT-MX or IOAM) or
+    locally (in case of INT-MX). The report might include INT-MD or IOAM
```
mickeyspiegel: s/INT-MX/INT-XD
I had it wrong in my previous comment

mickeyspiegel (approved):
One change required

Some non-technical cleanup is still required, as I noted in my Jan 14 comments. We can either do them in this PR or after it merges.

telemetry/specs/telemetry_report.mdk
rsivakolundu: I am ok with 4-octets each. Do we need a wider discussion?

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
-`
+      |                            Node id                            |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+      |RepType| InType| Report Length |   MD Length   |D|Q|F|I|  Rsvd |  |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  R
+      |            RepMdBits          |      Domain Specific ID       |  E
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  P
+      |            DSMdBits           |          DSMdstatus           |  O                        
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  R
+      |          Variable Optional Baseline & DS Metadata             |  T
```
RandyLevensalor, line 458: This seems to be overly constrained by not allowing for Truncated Packet or  Additional DS Extension Data in the same report.

Allowing for both within the same report will enable the full spectrum of data.  Plus it will allow an application that only knows about the packet fragment to read that and skip the DS extension data.

This will require that the "IntType" and Report Length would need to be duplicated for Domain Specific Extension Data and the packet fragment.

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,107 @@ The Destination Port is user configurable. The expectation is that the same
...
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+      |RepType| InType| Report Length |   MD Length   |D|Q|F|I|  Rsvd |  |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  R
+      |            RepMdBits          |      Domain Specific ID       |  E
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  P
+      |            DSMdBits           |          DSMdstatus           |  O                        
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  R
+      |          Variable Optional Baseline & DS Metadata             |  T                   
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  |
+      |      Truncated Packet or  Additional DS Extension Data        |  |
```
RandyLevensalor: This seems to be overly constrained by not allowing for Truncated Packet or Additional DS Extension Data in the same report.

Allowing for both within the same report will enable the full spectrum of data. Plus it will allow an application that only knows about the packet fragment to read that and skip the DS extension data.

This will require that the "IntType" and Report Length would need to be duplicated for Domain Specific Extension Data and the packet fragment.

RandyLevensalor (changes requested):
Including an option for both packet fragments and Additional DS Extension Data in the same report will make this a more complete solution.

In particular, I could see have a bit of DS Extension Data in a drop report with a packet fragment.  The DS extension can provide information on the system regarding the current roles on the system and the packet fragment which was dropped.

rsivakolundu:
We have 16 bits for DS Instructions for 4-byte wide DS metadata. Along with
baseline Instruction bits, we have an upper limit of 1024 bytes for
metadata (baseline and DS) and the packet fragment.  DS Instruction bits
are totally left up to the network management entity to define the kind of
metadata requested.

Do you envision more than 1024 bytes in a report per flow?

-Ramesh

RandyLevensalor:
I don't envision more than 1024 bytes in an individual flow report.

mickeyspiegel:
@RandyLevensalor Looking at INT 2.0, I don't see any restriction on the length represented by each DS Instruction bit other than a multiple of 4 bytes. If we relax this similarly in Telemetry Report 2.0, where one DS Instruction bit may represent more than 4 bytes of Variable Optional DS Metadata, would that be sufficient? Could you achieve what you want using DS Instruction bits?

RandyLevensalor:
@mickeyspiegel That would be fine.  Then we could remove   1: Domain Specific Extension Data from the InType, since that could be added to the DS metadata.

rsivakolundu:
I guess in that case, (MD Length - Baseline Instruction metadata length
calculated based upon bitmap) = DS Metadata. The collector needs to be
aware of the definition of DS Instruction Bitmap, which is anyways required.

I would like to keep the DS Extension Data encoding in the InType field.
Allows for a future extension when Truncated packet is not necessary for
the report.

Thanks

-Ramesh

mickeyspiegel:
@rsivakolundu +1 for keeping DS Extension Data in InType

RandyLevensalor:
That still limits the options if you need both a variable length (which is why I'm assuming you want to keep the DS Extension Data and not define everything in the bitmap.

That brings us back to having to choose or limit your DS data if you are including that packet fragment.

Note: This is a concern, not a strong objection.  It just feels like this is making assumptions about not needing the same level of DS data when the packet fragment is needed.

While this won't break my current use case, the assumption that you will not have the same DS requirements with and without packet fragments.

One assumption that I have is that multiple applications could read that same telemetry report.  Where some TR consumers are general-purpose and would ignore DS data. Then DS aware applications would handle the DS fields.

mickeyspiegel:
@RandyLevensalor Yes the options are a bit limited, but I don't see a particular use case that justifies throwing in yet another type and length. It is a judgement call.

Note that one use case for Additional DS Extension Data is non-packet specific report data, for example control plane information.

RandyLevensalor (approved):

RandyLevensalor:
I'd like to throw out one last suggestion before I drop this discussion around supporting both Domain Specific Extension Data and packet fragments.

To avoid adding extra data to the report header, could we create a new Inner Type for TLV reports.

The TLV would be of the structure of something like:
TVL - Type (Domain Specific Extension Data, IPv6, IPv4, Ethernet)
Length of the entry.
The actual data, either the packet fragment or Domain Specific Extension Data.

Followed by as many of these TLV entire.  All TLV entries and heders must be less than 1024 bytes to fit within the current report length.

rsivakolundu:
Are you proposing something like this?

|Ver = 2|   hw_id   |           Sequence Number                 |

|                           Node ID                             |

|RepType|In Type| Report Length |   MD Length   |D|Q|F|I|  Rsvd |

|            RepMdBits          |      Domain Specific          |

|            DSMdBits           |          DSMdstatus           |


|                 Variable Optional RepMdBits Data
|

|                                           Variable Optional DSMdBits
Data
|

|In Type|     TLV Length        |       Reserved        |NxtType|

 |                                                  Variable TLV Data
                   |

|In Type|     TLV Length        |       Reserved        |NxtType|

|                                                    Variable TLV Data
                   |

InType (4b): Inner Type – TLV formatted data or Truncated packet or
additional Domain Specific Data after Variable Optional Baseline & DS
Metadata corresponding to RepMdBits or DS MD Bits.

0 – None

1 – Domain Specific Extension Data

2 - Ethernet

3 – IPv4

4 – IPv6

5 - TLV

6-15 – Reserved.


NxtType (4b): Next Type – TLV formatted data or None after Variable TLV Data

0 – None

1 - TLV

2-15 – Reserved.


Example with DS Extension Data & Truncated Packet.


|Ver = 2|   hw_id   |           Sequence Number                 |

|                           Node ID                             |

|RepType|0 1 0 1| Report Length |   MD Length   |D|Q|F|I|  Rsvd |

|            RepMdBits          |      Domain Specific          |

|            DSMdBits           |          DSMdstatus           |


|                 Variable Optional RepMdBits Data              |


|                                           Variable Optional DSMdBits
Data
|

|0 0 0 1|      Length           |       Reserved        |0 0 0 1|

|                  Variable TLV Data (Domain Specific Extension Data)
    |


|0 1 0 0|      Length           |       Reserved        |0 0 0 0|

|                  Variable TLV Data (Truncated IPv4 Packet)
    |


If so, I think it works for any unforeseen use case.


-Ramesh

RandyLevensalor:
Ramesh,

Yes. That is exactly what I was thinking.

Thanks

mickeyspiegel:
That proposal is adding another type and length, which is something that I want to avoid. It is also mixing next header and TLV styles in a way that I do not feel brings out the best of either approach.

If necessary, we could revive the TLV format that I had proposed earlier in addition to the format that is currently in this PR. Personally, I don't see a strong enough rationale to add an additional format, but I won't object if others feel differently.

rsivakolundu:
I am fine with either approaches. Will go along with consensus opinions.

-Ramesh

mickeyspiegel:
@rsivakolundu The thing that made me uneasy more than anything else in the expanded format is "NxtType". That is what breaks from normal TLV style. I think it is unnecessary. If you drop that and just leave two reserved bytes, then InType 'TLV' means whatever follows MD Length is in TLV style. That includes the possibility of having multiple TLVs as long as they all fit within Report Length.

I can live with the expansion with that change. I would move the values around and make codepoint 1 TLV, with the understanding that in the TLVs the type value must be 2 or greater.

We probably would not support that codepoint initially in our implementation, but I don't see any issue with that.

mickeyspiegel:
Given that we don't have any flow identity information in the initial part of the report header, do we really need the InType codepoint "None"? Would there ever be a report without any additional information?

rsivakolundu:
I think so. Flow identification could be a domain specific metadata.

-Ramesh

mickeyspiegel:
@rsivakolundu Then just add the TLV codepoint as InType 1, shift the other codepoints from 1+ up 
by one, and define the TLV format as Type (4 bits), Reserved (4 bits), Length (8 bits), Reserved (16 bits), TLV contents (variable multiple of 4 bytes)

rsivakolundu:
We still need next Type field in the TLV.

InType (4b): Inner Type – TLV formatted data or Truncated packet or
additional Domain Specific Data after Variable Optional Baseline & DS
Metadata corresponding to RepMdBits or DS MD Bits.

0 – None

1 - TLV

2 – Domain Specific Extension Data

3 - Ethernet

4 – IPv4

5 – IPv6

6-15 – Reserved.

-Ramesh

mickeyspiegel:
@rsivakolundu I don't understand your comment about "next Type".

The first InType tells you what comes after the Variable Optional MdBits data.
When the value of that first InType is TLV, then there are one or more TLVs following the Variable Optional MdBits data, each a multiple of 4 bytes long. Each TLV includes a Type (from InType values 2+), a length, 2 reserved bytes, and a value. To determine whether another TLV is present, you have to evaluate the length of this and preceding TLVs to determine if Report Length has been filled up, as is normal for any TLV encoding scheme.

If you want to call the type in the TLVs something other than InType and copy InType codepoints 2+ in that definition, that might make things clearer.

rsivakolundu:
Agree with normal TLV encoding scheme. Since we have reserved bits, was
trying to simplify the processing at the collector.

Is the following acceptable?

Telemetry Report Header  (20+ Octets)



|Ver = 2|   hw_id   |           Sequence Number                 |

|                           Node ID                             |

|RepType|In Type| Report Length |   MD Length   |D|Q|F|I|  Rsvd |

|            RepMdBits          |      Domain Specific          |

|            DSMdBits           |          DSMdstatus           |


|                 Variable Optional RepMdBits Data
|

|                                           Variable Optional DSMdBits
Data
|

|TLVType|  Rsvd |   TLVLength   |           Reserved            |

|          Variable TLV Data as identified by TLVType           |


|TLVType|  Rsvd |   TLVLength   |           Reserved            |

|          Variable TLV Data as identified by TLVType           |

TLVType (4b): TLV Type – TLV data or Truncated packet or additional Domain
Specific Data after Variable Optional Baseline & DS Metadata corresponding
to RepMdBits or DS MD Bits.

0 – Domain Specific Extension Data

1 - Ethernet

2 – IPv4

3 – IPv6

4-15 – Reserved.



Report Length = MD Length + TLVLength + TLVLength in the above example.


-Ramesh

mickeyspiegel:
@rsivakolundu I am fine with the last proposal, except for two points of discussion:

1. Should the TLVType codepoint values align with the InType codepoint values? I guess I can go either way
2. In INT the shim header lengths never included the shim header itself (first 4 bytes). This was due to precedents set in other specifications such as VXLAN-GPE. For Telemetry Report, we need to decide whether the first 4 bytes are or are not included in the length. In Report Length, the first 4 bytes are just part of 12 fixed header bytes, so I guess including everything is the way I am leaning, but I am not very sure of myself. In TLVType there is a clear first 4 bytes. Should the first 4 bytes be included in the length?

For the Domain Specific Extensions codepoint, you could use the 2 bytes after type and length rather than making them reserved, but then the encoding would differ between the non-TLV Domain Specific Extensions and the TLV Domain Specific Extensions.

rsivakolundu:
Great. We are converging.

I prefer the TLVType codepoint values to independent of InType codepoint
values. Less confusion and allows for future use cases.

Including the first 4-bytes is more meaningful to me. I didn't like the
VXLAN-GPE precedence, but it is what it is at this stage. Nothing forces us
to not include the first 4-bytes.

I am leaning towards leaving the Domain Specific Extension to be free form
and not bitmapped. I could change those 16 bits to be Domain Specific
Extension Template.

Thoughts?

-Ramesh

mickeyspiegel:
@rsivakolundu Not sure what you mean by bitmapped. The first 16 bits (type, reserved, length) of each TLV must be the same. After that it can be anything you want. For truncated packet fragments, the next 16 bits should be reserved, but for other type values it should be free form depending on the type.

rsivakolundu:
Yes. I am stating the same thing. See description below.

TLVType (4b): TLV Type – TLV data or Truncated packet or additional Domain
Specific Data after Variable Optional Baseline & DS Metadata corresponding
to RepMdBits or DS MD Bits.

0 – Domain Specific Extension Data

1 - Ethernet

2 – IPv4

3 – IPv6

4-15 – Reserved.


Rsvd (4b) – Reserved. Should be set to zero upon transmission and ignored
upon reception.


TLVLength (8b) - Indicates the length, in 4-byte word, of Variable TLV Data
as identified by TLVType.


TLV Data Template (16b) – TLV Data Template specifies the format of the
Variable TLV Data. A non-zero TLV Data Template value specifies the
template for TLVType codepoint of Domain Specific Extension Data. The zero
TLV Data Template value is reserved for TLVType codepoints for Ethernet,
IPv4, IPv6.


Variable TLV Data – Variable length data based upon TLVType.


-Ramesh

mickeyspiegel:
@ramesh I would not even define the next two bytes as a general field. If you do, the last sentence needs to be reworded: "For TLVType codepoints Ethernet, IPv4, and IPv6, the TLV Data Template value should be zero upon transmission and ignored upon reception."

Instead, I would define a format for TLVType codepoints Ethernet, IPv4, and IPv6:
```
                               |       Reserved (16 bits)       |
|                        Truncated Packet                       |
```
Somewhere it needs to be stated that each TLV must be a multiple of 4 bytes long. If you stick with your proposed format, that would be in Variable TLV Data.

rsivakolundu:
TLVType (4b): TLV Type – TLV data or Truncated packet or additional Domain
Specific Data after Variable Optional Baseline & DS Metadata corresponding
to RepMdBits or DS MD Bits.

0 – Domain Specific Extension Data

1 - Ethernet

2 – IPv4

3 – IPv6

4-15 – Reserved.

Rsvd (4b) – Reserved. Should be set to zero upon transmission and ignored
upon reception.


TLVLength (8b) - Indicates the length, in 4-byte word, of Variable TLV Data
as identified by TLVType.


TLV Data Template (16b) – TLV Data Template specifies the format of the
Variable TLV Data. A non-zero TLV Data Template value specifies the
template for TLVType codepoint of Domain Specific Extension Data. For
TLVType codepoints Ethernet, IPv4, and IPv6, the TLV Data Template value
should be zero upon transmission and ignored upon reception.


Variable TLV Data – Variable length data, based upon TLVType, as 4-byte
words.


If we concur, I will add this format to the PR.


Thanks


-Ramesh

mickeyspiegel:
@rsivakolundu Go ahead and add it.

rsivakolundu:
Updated the PR with InType of TLV.

RandyLevensalor:
Looks great.  Thanks for making these changes!!

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,108 @@ The Destination Port is user configurable. The expectation is that the same
...
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+      
+* Ver (4b): Version
+  : This specification defines **version 2.0**.
+
+* hw_id (6b): Hardware ID
+  : Identifies the hardware subsystem within the network device that generated
+    this report. For example, in a chassis with multiple linecards this could
+    identify a specific linecard, or a subsystem within a linecard. The hw_id
+    is unique within the scope of a Switch id.
```
jklr: :s/Switch id/Node id

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,108 @@ The Destination Port is user configurable. The expectation is that the same
...
+  : Reflects the sequence of reports from a specific combination of
+    (Node id, hw_id) to a particular telemetry report destination. This
+    can be used to detect loss of telemetry reports before they reach their
+    intended destination.
 
-* Ver: Version
-  : This specification defines **version 1**.
+* Node ID (32b): Node ID
+  : The unique ID of a network device (such as routers, switches, NICs etc.) This is generally administratively assigned. Node IDs
+    must be unique within a management domain.
```
jklr, line 480: :s/management domain/INT domain.

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,108 @@ The Destination Port is user configurable. The expectation is that the same
...
 
-* Length
-  : Indicates the length of the telemetry report header in multiples of
-    4 octets, including the variable optional metadata.
+* RepType (4b): Report Type
+  : Type of Telemetry report
+    * 0: INT v2.0 or later.
+    * 1: IOAM
+    * 2 – 15: Reserved
```
jklr, line 486: I might have missed a discussion on RepType. It feels we need some clarifications on this field. 

1. I think we exclude INT v1.0 or 0.5 to be used with Report sepc 2.0. Can we clarify that?
1. IOAM version doesn't matter because the embedded IOAM header will have its own version field?
1. Is RepType the type of in-band telemetry embedded in the Truncated Packet? or does it matter even when there is no Truncated Packet?

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,108 @@ The Destination Port is user configurable. The expectation is that the same
...
-    * 2: IPv6
+* Q (1b): Congested Queue Association
+  : Indicates the presence of congestion on a monitored queue.
 
-* RepMdBits: Report Metadata Bits
+* F (1b): Tracked Flow Association
+  : Indicates that this telemetry report is for a tracked flow, i.e. the
+    packet matched a flow watchlist somewhere (in case of INT-MD, INT-MX or IOAM) or
+    locally (in case of INT-XD). The report might include INT-MD or IOAM
+    metadata beyond the inner ethernet header. Other telemetry reports are
```
jklr: Embedded Truncated Packet might not have inner ethernet depending on InType. Shall we say
"... metadata in the Truncated Packet."?

telemetry/specs/telemetry_report.mdk
```diff
@@ -424,47 +439,108 @@ The Destination Port is user configurable. The expectation is that the same
...
+    likely to be received for the same tracked flow, from the same network
+    device and (in case of drop reports, INT-MX, INT-XD or path changes) from
+    other network devices.
+  
+* I (1b): Intermediate Report
+  : Indicates that a transit node sent this intermediate report for INT-MD.
```
jklr, line 524: Can we provide a bit more context of this usecase?

telemetry/specs/telemetry_report.mdk
jklr: We use "INT domain" below when Domain Specific ID is explained.

jklr (changes requested):
Thanks for the great changes. Left a few minor comments.

telemetry/specs/telemetry_report.mdk
rsivakolundu: We agreed to make it multiple of 4-octet each in the email thread discussion.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Done.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Are you concluding that Node ID is unique only for a given Domain Specific ID. I am envisioning multiple Domain Specific ID within in management domain and Node id is unique in the entire management domain.

telemetry/specs/telemetry_report.mdk
rsivakolundu: Will add text with respect to INT v1.0 and v0.5 to clarify use of Report Spec v2.0
I would say it is still TBD. I can change it IOAM v(TBD).
RepType identifies if the Telemetry report is for INT v2.0 or for any other Telemetry.  It does not matter if a Truncated packet is there or not.

telemetry/specs/telemetry_report.mdk
rsivakolundu: I am ok with it.

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Bytes 4-11 after RepType are intentionally aligned in format with bytes 4-11 of INT-MX / INT-MD headers, so they can just be copied over when necessary.

We are hoping to take the same approach with IOAM, changing the format of bytes 4-11 to align with the IOAM Trace format, including all the IOAM bit definitions. This will make it easier to support IOAM Direct Export without having to remap instruction bits.

telemetry/specs/telemetry_report.mdk
jklr: That makes sense. 
As a separate note, we may need to rephrase the definition of INT domain in the INT spec. We can discuss in the next meeting.

telemetry/specs/telemetry_report.mdk
jklr: I see, that's neat.
Then it sounds like each RepType value should map to a version of each unique INT or IOAM header.
is 4b RepType large enough for that? 
Or the mapping of RepType value to INT/IOAM version can change when Telemetry Report spec version changes?

rsivakolundu: IOAM with Namespace extensions, similar to INT Domain Specific Extensions don't really map directly the way INT header bytes 4-11 are defined.  Namespace specific Trace Type and Namespace Specific Flags are not part of IOAM Header.  It is going to much harder to get the Namespace Extensions fields as part of default IOAM header.

rsivakolundu: I think mapping of RepType value to INT/IOAM version can change with Telemetry Report Spec version changes.

mickeyspiegel: @jklr I see 3 major enhancements in Telemetry Report 2.0 compared to Telemetry Report 1.0:

1. Support for domain specific extensions
2. Some new RepMdBit values
3. Coalescing of multiple reports in one packet

If supporting INT 1.0 or INT 0.5, there are no domain specific extensions. It looks like one RepMdBit value is missing in Telemetry Report 1.0 compared to INT 1.0, which is Level 2 Ingress Port ID + Egress Port ID (4 bytes each). I guess we did not want to go to a mix of 4 and 8 bytes per RepMdBit. Unless we want to cut a quick Telemetry Report 1.1 before merging this PR, that will be a drawback.

If supporting INT 1.0 or 0.5 and coalescing or level 2 port IDs are required, using the "INT 2.0" RepType looks like it would work. Perhaps we should just call it "INT" rather than "INT 2.0", even though it was designed with INT 2.0 in mind.

mickeyspiegel: @rsivakolundu Since IOAM does not have Namespace specific Trace Type and Namespace Specific Flags, where would the values come from?

If they are needed, it would have to be done through Domain Specific Extensions or equivalent, in a TLV if the truncated packet fragment is also required.

rsivakolundu: It is your call.

rsivakolundu: It is your call for Telemetry Report 1.1.

rsivakolundu: As far as Namespace Extensions, the current thinking is to trigger of Namespace ID and have an extension to the IOAM header to carry Namespace Specific Instructions and Flags.

mickeyspiegel: @rsivakolundu The IOAM data draft is in between last calls. Are people thinking about making major changes just before the supposed final last call?
If not, when/how would the format be extended?

For IOAM direct export, given the way the current proposal determines whether flow ID and sequence number are present based on length, it seems rather unfriendly to future extensions.

mickeyspiegel: For Telemetry Report 2.0, the proposal for now is to reserve a RepType codepoint for IOAM, but the format that goes along with that is not yet defined. I guess we would cut Telemetry Report 2.1 when the IOAM formats get finalized.

rsivakolundu: Makes sense to cut Telemetry Report 2.1 when IOAM formats are finalized.

telemetry/specs/telemetry_report.mdk
```diff
@@ -473,49 +549,138 @@ particular deployment.
...
 
     See the INT specification [^INT] for definitions of the remaining metadata.
+    
+* Domain Specific ID (16b)
+  : The unique ID of the INT Domain
+
+* DSMdBits (16b): Domain Specific Md Bits
+  : Bitmap that indicates which optional domain specific metadata is 
+    present in the record of the telemetry report. Each bit represents 
+    4 octets or multiples of 4 octets of domain specific optional metadata.
```
mickeyspiegel, line 560: When INT-MX sets a 'source-only' metadata bit, this introduces some complications for Telemetry Report 2.0 processing. At the source node, copying INT Hop ML into Telemetry Report MD Length will not work unless the resulting MD Length is increased to account for the 'source-only' metadata. At subsequent nodes, does that DSMdBit need to be cleared since that metadata will not be included in the telemetry report?

mickeyspiegel: I think the easiest thing to do is that the INT-MX source does not set the 'source-only' metadata bit in the data packet, but sets the bit and adjusts MD Length in the corresponding Telemetry Report 2.0 packet.

rsivakolundu: I agree that source node shouldn't set the "Source Only" bit for INT-MX. Instead should set the "source only" bit in DSMdBits and adjusts the MD Length. I will add some text to capture this behavior.

rsivakolundu: We should capture this INT-MX PR and here.

jklr: +1 for restraining source node from using any source-only in INT-MX header but doing so in the Report 2.0 packet.

telemetry/specs/telemetry_report.mdk
mickeyspiegel: I suggested some text for INT 2.1

mickeyspiegel:
@jklr Please revise comments or approve

jklr (approved):

telemetry/specs/telemetry_report.mdk
rsivakolundu: Will add some comments in INT 2.1 spec.

PR #63, opened 2019-12-05
revise INT over UDP

jklr:
that uses new UDP port number that will be allocated from IANA.

@rsivakolundu FYI.

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -326,8 +326,6 @@ This mode is similar to IFA [^IFA].
 * INT-PROBE-MD (eMbed Data): similar to INT-CLONE-MD, but INT is applied to
 special probe packets instead of clone packets.
 
-
-
 **NOTE: the rest of the spec is assuming INT-MD as the default mode, unless
 specified otherwise.**
 
@@ -672,8 +670,15 @@ cases:
     calculated using a hash of fields from the original packet, for example
     the original outer 5 tuple or the original L4 header fields.
     This is to enable a level of entropy for ECMP/LAG load balancing logic.
+    It is recommended that the checksum in the new UDP header be set to zero.
+    For IPv6 packets, this falls under the case of tunnel protocols,
+    which are allowed to use zero UDP checksums as specified in RFC 6936.
+    The existing L4 header will typically include a checksum computed
+    using the encapsulating IPv6 header fields, thus offering some protection
+    against IPv6 header corruption.
+
 In both cases, traffic with INT headers is likely to be hashed
-to a different path in the network as the UDP
+to a different path in the network as the new UDP
 destination port (INT_TBD) becomes part of the outer 5 tuple used by ECMP.
 
 The INT shim header for UDP has a field NPT (Next Protocol Type) that indicates
@@ -711,6 +716,7 @@ is optional.
     into the INT domain and there is no INT header present, it must ensure that
     the DSCP/Traffic class value is not the same as any of the values used
     to indicate INT.
+
 * Probe Marker fields: If DSCP field or values cannot be reserved for INT,
 probe marker option could be used. A specific 64-bit value can be inserted
 after the TCP/UDP header to indicate the existence of INT after TCP/UDP.
@@ -776,11 +782,11 @@ shim header in 4-byte words. A non-INT device may read this field
 and skip over INT headers.
 * UDP port, IP proto, or DSCP (16b): The contents of this field differ depending on the
 value of NPT.
-** NPT 0: The first byte and the last two bits of this 16b field are reserved,
+** NPT=0: The first byte and the last two bits of this 16b field are reserved,
 set to zero upon transmission and ignored upon reception. The first 6 bits of
 the second byte may optionally carry the original DSCP value.
-** NPT 1: The original UDP destination port value.
-** NPT 2: The first byte is reserved, set to zero upon transmission and ignored
+** NPT=1: The original UDP destination port value.
+** NPT=2: The first byte is reserved, set to zero upon transmission and ignored
 upon reception. The second byte carries the original IP protocol value.
 
 The other bits in the shim header are reserved (R) for future use,
@@ -1172,12 +1178,12 @@ TCP Header:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
-INT Shim Header for TCP/UDP, INT type is hop-by-hop:
+INT Shim Header for TCP/UDP, INT type is INT-MD, NPT is zero:
 `
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |     Type=1    |   Reserved    |  Length = 8   |   DSCP    |R R|
+   |Type=1 | 0 |R R|   Length=7    |  Reserved     |   DSCP    |R R|
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -777,9 +777,10 @@ and the last two bytes of the shim header carry the original UDP destination por
 ** two (2) indicates that another (the original) L4 header follows
 the INT stack, and the last byte of the shim header carries the IP protocol
 value for the L4 layer.
-* Length (8b): This is the total length of INT metadata header, INT stack and the
-shim header in 4-byte words. A non-INT device may read this field
-and skip over INT headers.
+* Length (8b): This is the total length of INT metadata header and INT stack
+in 4-byte words. The length of the shim header (1 word) is NOT counted
+since INT version 2.0.
+A non-INT device may read this field and skip over INT headers.
 * UDP port, IP proto, or DSCP (16b): The contents of this field differ depending on the
 value of NPT.
 ** NPT=0: The first byte and the last two bits of this 16b field are reserved,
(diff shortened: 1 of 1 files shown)


telemetry/specs/INT.mdk
```diff
@@ -652,8 +652,32 @@ both INT header stacks carry information for respective layers and need not be
...
+    The source port number of the new UDP header must be either 1) copied from the
+    original L4 header, 2) computed as a hash of the existing L4 header
+    fields (such as source/destination ports in case of TCP/UDP), or 3)
+    computed as a hash of the original outer 5 tuple. The second option
+    is recommended to provide ECMP/LAG load-balancing logic with the same degree of
+    entropy of the original data packet.
+The INT shim header has a field NPT (Next Protocol Type) that indicates
+which of the two cases are applied to a given INT packet. In case a new UDP
+header was inserted, the INT sink will strip the UDP header off from the
+packet together with the rest of INT headers.
```
mhira1: We should explicitly state the complete behavior that is expected from the INT sink.
For the case that original packet already had UDP header, INT sink must copy original destination port number into UDP header field and strip the INT header.
For the case that UDP header was added to the packet, INT sink must copy original protocol type to IP header, and strip UDP + INT headers

telemetry/specs/INT.mdk
```diff
@@ -652,8 +652,32 @@ both INT header stacks carry information for respective layers and need not be
 considered interfering with each other.
 
 A field in Ethernet, IP, or TCP/UDP should indicate if the
-INT header exists after the TCP/UDP header. We propose two options.
-
+INT header exists after the TCP/UDP header. We propose three options.
+
+* UDP destination port field: a new UDP port number (INT_TBD) will be assigned
+by IANA to indicate the existence of INT after UDP. This option supports two
+use cases:
```
mhira1: In both cases, traffic with INT headers is likely to be hashed to a different path in the network in ECMP scenarios.

telemetry/specs/INT.mdk
```diff
@@ -724,22 +748,32 @@ INT shim header for TCP/UDP:
...
 the destination header type (See Section [#sec-int-header-types]).
-* Length: This is the total length of INT metadata header, INT stack and the
+* NPT (Next Protocol Type, 2b): This field is meaningful only when the UDP destination
+port number (INT_TBD) is used to indicate the existence of INT. In the other cases,
+this field must be zero. When UDP destination port is INT_TBD, this field may have
+one of the two values:
+** one (1) indicates that UDP payload is followed by the INT stack, and
+the last two bytes of the shim header carries the UDP destination port.
+** two (2) indicates that another (the original) L4 header is followed by
+the INT stack, and the last byte of the shim heaer carries the IP protocol
```
mhira1: typo: "shim heaer" -> "shim header"

telemetry/specs/INT.mdk
```diff
@@ -724,22 +748,32 @@ INT shim header for TCP/UDP:
...
-* Type: This field indicates the type of INT Header following the shim header.
+* Type (4b): This field indicates the type of INT Header following the shim header.
 Two Type values are used: one for the hop-by-hop header type and the other for
 the destination header type (See Section [#sec-int-header-types]).
-* Length: This is the total length of INT metadata header, INT stack and the
+* NPT (Next Protocol Type, 2b): This field is meaningful only when the UDP destination
+port number (INT_TBD) is used to indicate the existence of INT. In the other cases,
+this field must be zero. When UDP destination port is INT_TBD, this field may have
+one of the two values:
+** one (1) indicates that UDP payload is followed by the INT stack, and
```
mhira1: Some clarification of where the INT stack and INT shim header is located for different cases would be useful. Previously INT stack and INT shim header was after TCP/UDP header. This revision is suggesting INT stack is after UDP payload. Is that the case for INT natively in TCP also?

jklr: > INT stack is after UDP payload

no that's not the message. I intended the original UDP payload follows the INT stack, not the other way around. Was the writing confusing?

jklr:
@mhira1 can you pls review the revision and approve this PR? Thx!

telemetry/specs/INT.mdk
```diff
@@ -652,7 +652,40 @@ both INT header stacks carry information for respective layers and need not be
...
+
+* UDP destination port field: a new UDP port number (INT_TBD) will be assigned
+by IANA to indicate the existence of INT after UDP. This option supports two
+use cases:
+  - The original packet already has UDP header either as user application
+    protocol or as part of another UDP-based encapsulation such as VXLAN,
+    GENEVE, RoCEv2. INT is inserted after the UDP header with the UDP
+    destination port number changed to INT_TBD.
+    The original destination port number
+    is carried in the shim header for the INT sink to restore back when it
```
mickeyspiegel: nit: s / restore back when / restore, when

telemetry/specs/INT.mdk
```diff
@@ -652,7 +652,40 @@ both INT header stacks carry information for respective layers and need not be
...
+  - A new UDP header for INT is inserted between IP and the existing L4 header.
+    The protocol field of IP header is set to 17 for UDP and the original
+    IP protocol value is carried in the INT shim header.
+    In the new UDP header, INT_TBD is used as the destination port number.
+    The source port number of the new UDP header must be either 1) copied from the
+    original L4 header, 2) computed as a hash of the existing L4 header
+    fields (such as source/destination ports in case of TCP/UDP), or 3)
+    computed as a hash of the original outer 5 tuple. The second option
+    is recommended to provide ECMP/LAG load-balancing logic with the same degree of
+    entropy of the original data packet.
```
mickeyspiegel: I prefer to weaken this language a little, leaving flexibility for switches to do entropy however they want. Since this case keeps the existing L4 header intact, we should probably treat this like we do any tunnel. Normally this would be along the lines of option 3, but many switches offer the user configurability of hash functions including which fields to include, what the hash function should be, etc.

Suggested replacement text for UDP source port:
It is recommended that the source port number of the new UDP header be calculated using a hash of fields from the original packet, for example the original outer 5 tuple. This is to enable a level of entropy for ECMP/LAG load balancing logic.

mickeyspiegel: We should also add some language about the UDP checksum:

It is recommended that the checksum in the new UDP header be set to zero. For IPv6 packets, this falls under the case of tunnel protocols, which are allowed to use zero UDP checksums as specified in RFC 6935. The existing L4 header will typically include a checksum computed using the encapsulating IPv6 header fields, thus offering some protection against IPv6 header corruption.

telemetry/specs/INT.mdk
mickeyspiegel: Text was added below.

telemetry/specs/INT.mdk
```diff
@@ -724,22 +757,31 @@ INT shim header for TCP/UDP:
...
+* Type (4b): This field indicates the type of INT Header following the shim header.
 Two Type values are used: one for the hop-by-hop header type and the other for
 the destination header type (See Section [#sec-int-header-types]).
-* Length: This is the total length of INT metadata header, INT stack and the
+* NPT (Next Protocol Type, 2b): This field is meaningful only when the UDP destination
+port number (INT_TBD) is used to indicate the existence of INT. In the other cases,
+this field must be zero. When UDP destination port is INT_TBD, this field may have
+one of the two values:
+** one (1) indicates that the original UDP payload is followed by the INT stack,
+and the last two bytes of the shim header carries the original UDP destination port.
```
mickeyspiegel: nit: s/carries/carry

telemetry/specs/INT.mdk
```diff
@@ -724,22 +757,31 @@ INT shim header for TCP/UDP:
...
+** two (2) indicates that another (the original) L4 header is followed by
```
mickeyspiegel: This sounds backwards. My assumption is that the encap is:
outer ethernet
outer IP
new UDP
INT
original L4 header
payload

s / is followed by / follows

telemetry/specs/INT.mdk
```diff
@@ -724,22 +757,31 @@ INT shim header for TCP/UDP:
...
+* Length (8b): This is the total length of INT metadata header, INT stack and the
 shim header in 4-byte words. A non-INT device may read this field
 and skip over INT headers.
-
-* DSCP: If IP DSCP is used to indicate INT, this field optionally
-stores the original DSCP value. Otherwise, this field
+* UDP port, IP proto, or DSCP: If IP DSCP is used to indicate INT, this field optionally
+stores the original DSCP value. For non-zero values of NPT, either UDP port number
+or IP protocol number is carried accordingly. Otherwise, this field
 is reserved.
```
mickeyspiegel: This text is not so clear. I prefer to spell it out:

UDP port, IP proto, or DSCP: The contents of this field differ depending on the value of NTP:
NTP 0: The first byte and the last two bits are reserved, set to zero upon transmission and ignored upon reception. The second byte may optionally carry the original DSCP value.
NTP 1: The original UDP destination port value.
NTP 2: The first byte is reserved, set to zero upon transmission and ignored upon reception. The second byte carries the original IP protocol value.

telemetry/specs/INT.mdk
```diff
@@ -724,22 +757,31 @@ INT shim header for TCP/UDP:
...
 
-All the other bits in the shim header are reserved for future use.
+The other bits in the shim header are reserved (R) for future use.
```
mickeyspiegel: , set to zero upon transmission and ignored upon reception.

mickeyspiegel (changes requested):

telemetry/specs/INT.mdk
jklr: done

telemetry/specs/INT.mdk
jklr: @mickeyspiegel I addressed the first comment. As for the UDP checksum, do we need additional language on top of the existing texts under "Checksum Update" section? Line 561-.

telemetry/specs/INT.mdk
jklr: good catch. Fixed.

telemetry/specs/INT.mdk
jklr: > NTP 0: ... The second byte may optionally carry the original DSCP value.

To be precise, did you mean?
> The first 6 bits of the second byte may optionally carry the original DSCP value.

mickeyspiegel: Yes

telemetry/specs/INT.mdk
mickeyspiegel: That section is about checksum update, when the received packet included a checksum. In that case, if the received UDP checksum is zero, then there is no checksum update. If the received UDP checksum is non-zero, then the checksum must be updated in one of the ways specified in that section.

This case differs a little, since in this case we are adding the UDP header. In this case, the options are to either insert a UDP checksum with value 0, or compute a valid UDP checksum.

I don't care whether we cover this here, or add some text to section 5.5, but it should be discussed somewhere.

mickeyspiegel (approved):

mickeyspiegel:
Should we add an example in Section 6?

telemetry/specs/INT.mdk
jklr: Okay my original text was confusing. Fixed them. Thx.

telemetry/specs/INT.mdk
jklr: Added some text here.

jklr:
Will add a new set of example for this UDP encap, once #54 is merged.

PR #65, opened 2020-01-16
Increase ingress and egress timestamp size to 8 bytes each for telemetry report spec 2.0

mickeyspiegel:


jklr:
@mickeyspiegel let's aim to revise this PR before the next meeting. Thanks!

mickeyspiegel:
Already addressed in separate pull requests for INT 2.0 and Telemetry Report 2.0.

PR #66, opened 2020-01-25
simplify INT modes diagram

jklr:
remove INT-CLONE-MD and INT-PROBE-MD.


jklr:
@orr101 as you suggested.

PR #67, opened 2020-02-05
New example for INT over UDP. Source-only DS metadata

jklr:
This PR also
1. change the term hop-by-hop to INT-MD
1. don't allow transit devices to set 'D' (discard) bit. It's only for the source to set.
1. allow 'source-only' domain specific metadata and relax INT length requirements

Need to discuss/decide
- any restrictions on the source-only instruction bits and semantics
- keep or drop INT "Destination type"?

Once we agree on the details of 'source-only' DS metadata, we will need to update the INT modes descriptions such that INT-MD and MX can also support edge-to-edge usecase via the source-only metadata.

@orr101 @rsivakolundu for info.

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -1,6 +1,6 @@
-Title         : In-band Network Telemetry (INT) Version 2.0 Dataplane Specification
-Title Note : Working draft. Note: consider using tagged versions for implementation.
-Title Footer: 2018-08-17
+Title         : In-band Network Telemetry (INT) Dataplane Specification
+Title Note : This is a working draft towards v2.0. Consider using tagged versions for implementation.
+Title Footer: 2020-02-04
 Author : The P4.org Applications Working Group. Contributions from
 
 Affiliation : *Alibaba, Arista, Barefoot Networks, Cisco Systems, Dell, Intel, Marvell, Netronome, VMware*
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -289,7 +289,6 @@ at their Flow Watchlists. No packet Modification is needed. This mode was also
 known as "Postcard" mode in the previous versions of Telemetry Report spec,
 originally inspired by [^Postcard].
 
-
 * **INT-MX** (eMbed instruct(X)ions): INT Source devices embed INT instructions in
 the packet header, each following Source/Transit device directly sends the metadata
 to the monitoring system by following the instructions embedded in the packets.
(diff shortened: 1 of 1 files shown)


telemetry/specs/INT.mdk
```diff
@@ -1036,20 +1039,28 @@ each bit corresponds to a specific standard metadata as specified in Section 3.
...
-  * Domain Specific ID (16b): the unique ID of the INT Domain.
+* Domain Specific ID (16b): the unique ID of the INT Domain.
   
-  * DS Instruction (16b): Instruction bit map specific to the INT domain identified by the 
+* DS Instruction (16b): Instruction bit map specific to the INT domain identified by the 
   Domain Specific ID. Domain Specific Instruction is an instruction that requires additional 
   processing of Domain Specific Flags (DS Flags) for the INT Domain identified by Domain Specific ID. 
+
+  Some instruction bits can be defined as 'source-only' metadata by the INT domain. Those metadata
+  will be inserted only by the INT source, not by INT transit or sink devices.
```
RandyLevensalor, line 1063: This is great.  Thanks for adding it!

telemetry/specs/INT.mdk
```diff
@@ -1088,28 +1106,25 @@ header.
     a version 1.0 INT switch cannot operate alongside version 2.0 INT switches
     if a new 8B metadata is introduced in version 2.0, as the version 1.0
     INT switch could insert 0xFFFFFFFF reserved value for a 8B metadata field,
-    thus breaking the metadata stack length invariance - the length of
-    metadata stack will not be a multiple of Per-Hop Metadata length \* 4
-    in this case.
+    thus breaking the metadata length invariance - the length of
+    metadata stack (excluding 'source-only' Domain Specific Metadata)
+    must be a multiple of Hop ML \* 4.
```
mickeyspiegel: I think most of the text above is no longer relevant. Instead, you have to follow the rules at 1059 above. It does not matter whether the unknown bits are from the base instruction bitmap or the DS instructions, either way you have to pad out to Hop ML.

mickeyspiegel (approved):
One comment that can be addressed in this PR or separately.

telemetry/specs/INT.mdk
jklr: Removed lines 1109-1111. 
Kept the text about the major revision requirement for adding non-4B metadata. The padding out to Hop ML must honor the order and the size of each metadata.

telemetry/specs/INT.mdk
jklr: Thanks for your timely inputs from your usecase!

PR #71, opened 2020-02-13
Correction to the fixed header length in INT.mdk

### LINKED ISSUE #70: Update fixed header size in INT.mdk
The fixed header size used to be 12 bytes and with the addition of Domain Specific ID (2 bytes) ad Domain Specific Instruction (2 bytes), the new fixed header size is 12+4= 16 bytes. 
Section 5.7 in INT.mdk mentions the INT fixed header size to be 12 bytes and would need to be updated.

Sanjhana-Jayagopal:
Fixes #70 
With the addition of Domain Specific ID (2 bytes) ad Domain Specific Instruction (2 bytes), the new fixed header size is 16 bytes. This PR updates the INT fixed header size in Section 5.7 of INT.mdk.

### DIFF
--- a/telemetry/specs/INT.mdk
+++ b/telemetry/specs/INT.mdk
@@ -1139,7 +1139,7 @@ the INT metadata header.
 * The length (in bytes) of the INT metadata stack must always
 be a multiple of (Hop ML \* 4), plus the size of 'source-only' Domain Specific
 Metadata if added by the source. The total stack length can be determined
-by subtracting the total INT fixed header sizes (12 bytes)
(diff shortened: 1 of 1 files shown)


RandyLevensalor (approved):
Good catch.

jklr:
@Sanjhana-Jayagopal 
Thanks for the review!
We've recently changed the meaning of shim header length. It used to include itself (4B) but not any more, canceling out the additional 4B added for Domain extension. Hence 12B is correct.

We much appreciate any review comments on the draft. We will cut v2.0 tomorrow, and the draft has many recent changes, likely to have some confusing or incorrect statements.

PR #72, opened 2020-02-14
some minor clarification edits

js9045:
sorry, its not merged to latest master but you should be able to pick up the suggestions from the changed lines (not the removed ones).


jklr:
Manually merged into master. Thanks.

PR #77, opened 2020-04-06
INT-MX details

rsivakolundu:


telemetry/specs/INT.mdk
```diff
@@ -295,7 +295,9 @@ INT Sink devices strip the instruction header before forwarding the packet to th
 receiver.
 Packet modification is limited to the instruction header, the packet size doesn't
 grow as the packet traverses more Transit devices.
-This mode is inspired by IOAM's "Immediate Export" [^IOAM].
+
+This mode is inspired by IOAM's "Direct Export" [^IOAM].
+
 A new INT type header for INT-MX will be defined in the next minor revision.
```
mickeyspiegel: Delete this sentence since the INT-MX header is being defined in this version.

telemetry/specs/INT.mdk
```diff
@@ -295,7 +295,9 @@ INT Sink devices strip the instruction header before forwarding the packet to th
 receiver.
```
mickeyspiegel: nit: Proposed rewording of this paragraph:

The INT Source device embeds INT instructions in
the packet header, then the INT Source, each INT Transit, and the INT sink directly send the
metadata to the monitoring system by following the instructions embedded in the packets.
The INT Sink device strips the instruction header before forwarding the packet to the
receiver.

telemetry/specs/INT.mdk
```diff
@@ -295,7 +295,9 @@ INT Sink devices strip the instruction header before forwarding the packet to th
...
 Packet modification is limited to the instruction header, the packet size doesn't
 grow as the packet traverses more Transit devices.
-This mode is inspired by IOAM's "Immediate Export" [^IOAM].
+
+This mode is inspired by IOAM's "Direct Export" [^IOAM].
```
mickeyspiegel: IOAM Direct Export is a separate draft, so the reference should be modified.

telemetry/specs/INT.mdk
```diff
@@ -372,8 +374,8 @@ must be unique within an INT domain.
 
 ## Ingress Information
 
-* Ingress port identifier
-  - The port on which the INT packet was received. A packet may be received
+* Ingress interface identifier
+  - The interface on which the INT packet was received. A packet may be received
```
mickeyspiegel, line 376: The word "port" appears all over this paragraph. Except when referring to "physical port" or "link aggregation port group", these occurrences should be changed to "interface".

telemetry/specs/INT.mdk
```diff
@@ -396,7 +398,7 @@ must be unique within an INT domain.
 
 ## Egress Information
 
-* Egress port identifier
+* Egress interface identifier
```
mickeyspiegel, line 399: The word "port" appears all over this paragraph. Except when referring to "physical port" or "link aggregation port group", these occurrences should be changed to "interface".

telemetry/specs/INT.mdk
```diff
@@ -1674,8 +1937,8 @@ at the time packets were forwarded.
 
 ## Ingress
 
-* Ingress port identifier
-  - The port on which the INT packet was received. A packet may be received
+* Ingress interface identifier
+  - The interface on which the INT packet was received. A packet may be received
```
mickeyspiegel, line 2242: The word "port" appears all over this paragraph. Except when referring to "physical port" or "link aggregation port group", these occurrences should be changed to "interface".

telemetry/specs/INT.mdk
```diff
@@ -1705,8 +1968,8 @@ INT framework leaves those decisions to device vendors.
 
 ## Egress
 
-* Egress port identifier
-  - The port on which the INT packet was sent out. A packet may be transmitted
+* Egress interface identifier
+  - The interface on which the INT packet was sent out. A packet may be transmitted
```
mickeyspiegel, line 2273: The word "port" appears all over this paragraph. Except when referring to "physical port" or "link aggregation port group", these occurrences should be changed to "interface".

telemetry/specs/INT.mdk
```diff
@@ -1687,7 +1950,7 @@ at the time packets were forwarded.
 * Ingress timestamp
   - The device local time when the INT packet was received on the **_ingress_**
 physical or logical port.
-* Ingress port RX pkt count
+* Ingress interface RX pkt count
```
mickeyspiegel, line 2254: Either change all four "Ingress port RX" in this section to "Ingress interface RX", or leave them all as "Ingress port RX". Do the same for "Egress port TX" in the next section.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
```
mickeyspiegel, line 485: The last sentence just above needs to be modified to refer to the section defining the INT-MX format.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
 
 ## Per-Hop Header Creation
 
-In the INT model, each device in the packet forwarding path creates additional
-space in the INT Header on-demand to add its own INT metadata. To avoid
+### INT Source Node
+
+In the INT-MD, INT-MX model, the INT Source device in the packet forwarding
```
mickeyspiegel: nit: In the INT-MD and INT-MX modes, ...

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+In the INT-MD, INT-MX model, the INT Source device in the packet forwarding 
+path creates the INT-MD or INT-MX Header. 
+
+In INT-MD, the source node add its own INT metadata after the header. To avoid 
+exhausting header space in the case of a forwarding loop or any other anomalies, 
+it is strongly recommended to limit the number of total INT metadata fields added 
+by Transit Hop devices by setting the Remaining Hop Count field in INT header 
+appropriately. 
+
+The INT-MD and INT-MX header is described in detail in the subsequent
```
mickeyspiegel: nit: The INT-MD and INT-MX headers are ...

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
 
 ## Per-Hop Header Creation
```
mickeyspiegel: This section is no longer just about "Creation". "Operations"?

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+it is strongly recommended to limit the number of total INT metadata fields added 
+by Transit Hop devices by setting the Remaining Hop Count field in INT header 
+appropriately. 
+
+The INT-MD and INT-MX header is described in detail in the subsequent
+sections.
+
+### INT Transit Hop Node
+
+In the INT-MD model, each device in the packet forwarding path creates additional
```
mickeyspiegel: nit: s/model/mode

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+space in the INT-MD Header on-demand to add its own INT metadata. To avoid
 exhausting header space in the case of a forwarding loop or any other anomalies,
 it is strongly recommended to limit the number of total INT metadata fields
```
mickeyspiegel: All this strongly recommended stuff makes no sense for INT Transit Hops, so this sentence should be simplified:

To avoid
exhausting header space in the case of a forwarding loop or any other anomalies,
each INT Transit Hop must decrement the Remaining Hop Count field in the INT
header appropriately.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
-added by devices by setting the Remaining Hop Count field in INT header
+added by devices by decrementing the Remaining Hop Count field in INT header
 appropriately.
 
+In the INT-MX model, each device in the packet forwarding path follows the 
+intructions in the INT-MX Header, gathers the device specific metadata and 
+exports the device metadata using the Telemetry Report. The devices in the 
+packet forwarding path do not modify the INT-MX header fields.
```
mickeyspiegel: s/The devices in the packet forwarding path/INT Transit Hop devices

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+
+### INT Sink Node
+
+In the INT-MD and INT-MX model, the INT sink node device does not modify 
+or add its own INT metadata to the INT-MD/INT-MX Header.
```
mickeyspiegel: I don't think we should describe internal processing at the sink node. Actually for INT-MD I don't think it is correct, since an implementation could add its INT metadata to the INT-MD header, then export with the embedded INT metadata, then remove the INT-MD header.

Just delete this paragraph and let the next paragraphs take care of it.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+exports the device metadata using the Telemetry Report. The devices in the 
+packet forwarding path do not modify the INT-MX header fields.
+
+### INT Sink Node
+
+In the INT-MD and INT-MX model, the INT sink node device does not modify 
+or add its own INT metadata to the INT-MD/INT-MX Header. 
+
+In INT-MD model, INT Sink node removes the INT Headers, extracts the 
+INT Headers and Metadata stack and decides on sending the collected information.
```
mickeyspiegel: In INT-MD mode, the INT Sink node removes the INT Headers and Metadata stack from the packet, and decides whether to report the collected information.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+
+In the INT-MD and INT-MX model, the INT sink node device does not modify 
+or add its own INT metadata to the INT-MD/INT-MX Header. 
+
+In INT-MD model, INT Sink node removes the INT Headers, extracts the 
+INT Headers and Metadata stack and decides on sending the collected information. 
+
+In INT-MX model, INT Sink node removes the INT-MX header, gathers the 
+device specific metadata and exports the device metadata using the 
+Telemetry Report.
```
mickeyspiegel: In INT-MX mode, the INT Sink node removes the INT-MX header, gathers the device specific metadata, and decides whether to report that metadata.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+
+
 ## MTU Settings
 
-As each hop creates additional space in the INT header to add its metadata,
-the packet size increases. This can potentially cause egress link MTU to be
-exceeded at an INT switch.
+In INT-MD model, as each hop creates additional space in the INT header
```
mickeyspiegel: nit: In INT-MD mode, ...

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+
+
 ## MTU Settings
 
-As each hop creates additional space in the INT header to add its metadata,
-the packet size increases. This can potentially cause egress link MTU to be
-exceeded at an INT switch.
+In INT-MD model, as each hop creates additional space in the INT header 
+to add its metadata, the packet size increases. This can potentially 
+cause egress link MTU to be exceeded at an INT switch.
```
mickeyspiegel, line 531: In both INT-MX and INT-MD modes, it is possible that insertion of the INT header at the INT Source node may cause the egress link MTU to be exceeded.

telemetry/specs/INT.mdk
```diff
@@ -486,18 +488,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
 
 This may be addressed in the following ways -
```
mickeyspiegel, line 534: There is a reference to "INT metadata header length (8B)" which needs to be adjusted to the new 12B length in the paragraph below

telemetry/specs/INT.mdk
```diff
@@ -1114,12 +1150,12 @@ INT-MD Metadata Header and Metadata Stack:
 * INT instructions are encoded as a bitmap in the 16-bit INT Instruction field:
```
mickeyspiegel, line 1153: Just above this there is a discussion of what to do when receiving a packet with Remaining Hop Count 0. Since the I bit is being added to the telemetry report, do you want to describe that alternative processing? How would the node know what value to reset Remaining Hop Count to?

Alternatively, the device may report the received metadata stack, and reset the Remaining Hop Count.

mickeyspiegel (changes requested):
Mostly minor things.
Only reviewed up until the INT-MX definition so far, more to come.

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: I think I changed port to interface.

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Makes sense.

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Makes sense.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Added.

telemetry/specs/INT.mdk
rsivakolundu: Changed

telemetry/specs/INT.mdk
rsivakolundu: The "I" bit in the Telemetry Report is to handle MTU exceeded. Didn't consider using that for this situation. It is dangerous to continue processing INT if RemainingHopCnt is zero. It indicates that the source node didn't use the correct RemainingHopCnt based upon the number of hops in the INT domain. The assumption is that the network monitoring entity knows the topology and the max number of hops in the INT domain.  We can discuss this further in a WG meeting.

telemetry/specs/INT.mdk
mickeyspiegel: Not done

telemetry/specs/INT.mdk
```diff
@@ -484,20 +484,51 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+
+In INT-MD, the source node add its own INT metadata after the header. To avoid 
+exhausting header space in the case of a forwarding loop or any other anomalies, 
+it is strongly recommended to limit the number of total INT metadata fields added 
+by Transit Hop devices by setting the Remaining Hop Count field in INT header 
+appropriately. 
 
-In the INT model, each device in the packet forwarding path creates additional
-space in the INT Header on-demand to add its own INT metadata. To avoid
+The INT-MD and INT-MX header are described in detail in the subsequent
```
mickeyspiegel: nit: s/header/headers

telemetry/specs/INT.mdk
```diff
@@ -484,20 +484,51 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+sections.
+
+### INT Transit Hop Node
+
+In the INT-MD mode, each device in the packet forwarding path creates additional
+space in the INT-MD Header on-demand to add its own INT metadata. To avoid
```
mickeyspiegel, line 505: nit: Please delete white space at the end of the line

telemetry/specs/INT.mdk
```diff
@@ -1674,62 +1934,62 @@ at the time packets were forwarded.
...
-    on an arbitrary stack of port constructs starting with a physical port.
+* Ingress interface identifier
+  - The interface on which the INT packet was received. A packet may be received
+    on an arbitrary stack of interface constructs starting with a physical port.
     For example, a packet may be received on a physical port that belongs to a
     link aggregation port group, which in turn is part of a Layer 3 Switched
     Virtual Interface, and at Layer 3 the packet may be received in a tunnel.
-    Although the entire port stack may be monitored in theory, this
+    Although the entire interface stack may be monitored in theory, this
     specification allows for monitoring of up to two levels of ingress port
```
mickeyspiegel: s/port/interface

telemetry/specs/INT.mdk
```diff
@@ -1674,62 +1934,62 @@ at the time packets were forwarded.
...
-    on an arbitrary stack of port constructs ending at a physical port.
+* Egress interface identifier
+  - The interface on which the INT packet was sent out. A packet may be transmitted
+    on an arbitrary stack of interface constructs ending at a physical port.
     For example, a packet may be transmitted on a tunnel, out of a Layer 3
     Switched Virtual Interface, on a Link Aggregation Group, out of a
     particular physical port belonging to the Link Aggregation Group.
-    Although the entire port stack may be monitored in theory, this
+    Although the entire interface stack may be monitored in theory, this
     specification allows for monitoring of up to two levels of egress port
```
mickeyspiegel: s/port/interface

telemetry/specs/INT.mdk
```diff
@@ -1674,62 +1934,62 @@ at the time packets were forwarded.
...
-    identifiers. The semantics of port identifiers may differ across devices,
-    each INT hop chooses the port type it reports at each of the two levels.
+    identifiers. The semantics of interface identifiers may differ across devices,
+    each INT hop chooses the interface type it reports at each of the two levels.
 * Egress timestamp
   - The device local time when the INT packet was processed by the egress
-physical or logical port.
-* Egress port TX pkt count
+physical port or logical interface.
+* Egress interace TX pkt count
```
mickeyspiegel: typo: s/interace/interface

telemetry/specs/INT.mdk
mickeyspiegel: Then wouldn't it make sense to add this alternative processing option to the MTU section above?

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
 by subtracting the total INT fixed header sizes (12 bytes)
 from (shim header length \* 4).
 
+## INT-MX Metadata Header Format
```
mickeyspiegel: While the INT-MX Header includes bits requesting reporting of certain metadata, there is no metadata in the header itself. Should it just be "INT-MX Header Format"?

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+monitoring system in the Telemetry Report.
+
+Details of the metadata semantics and format of the Telemetry Report can 
+be accessed at the link below:
+  https://github.com/p4lang/p4-applications/blob/master/docs/telemetry_report_v2.0.pdf
+
+* The fields in the INT metadata header are interpreted the following way:
+  - Ver (4b): INT metadata header version. Should be 2 for this version.
+  - D   (1b): Discard Copy/Clone. Not applicable in INT-MX mode. Should be set to 0 
+         by the INT Source.
```
mickeyspiegel: While a copy or clone does not make any sense for INT-MX, a probe might make sense. Do we really want to get rid of this bit?

Perhaps we should just call it "Discard" without "Copy/Clone", even in the INT-MD section above.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  - E   (1b): Max Hop Count exceeded. Not applicable in INT-MX mode. Should be set to 0 
+         by the INT Source.
```
mickeyspiegel: Not sure if we should define the bit explicitly or just leave it as part of the reserved bits.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  - M   (1b): MTU exceeded. Not applicable in INT-MX mode. Should be set to 0 
+         by the INT Source.
```
mickeyspiegel: Not sure if we should define the bit explicitly or just leave it as part of the reserved bits.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  - R  (25b): Reserved bits. Should be set to 0 by INT Source and ignored by other devices.
```
mickeyspiegel: While originally I thought that Hop ML / MD Length is irrelevant in INT-MX, with the latest Telemetry Report format we do have an MD Length that indicates where the inner contents of the Report Header begin. That suggests that MD Length should be present here in the third byte of the INT-MX header, from where it can be directly copied into the Telemetry Report header. That would leave some reserved bits after the D (or M if we keep the E and M definitions) bit, and before the MD Length which could be anywhere between 5 and 8 bits in length. Then another 8 bits of reserved after that.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  will be sent only by the INT source, not by INT transit or sink devices.
+  In a sense, 'source-only' bits do not serve as instructions for downstream INT devices
+  to follow. In INT-MX mode, the source node should not set the bits corresponding to 
+  "source-only" metadata in the DS Instruction bitmap. However, the source node should ensure
+  that it sets the bits corresponding to "source-only" metadata in DSMdBits bitmap of the 
+  Telemetry Report. This is needed for the monitoring system (or any consumer of the metadata) 
+  to know how to parse and use the data.
+
+  The amount of Domain Specific Metadata sent by each hop must be a multiple of 4 bytes,
+  determined from the Domain Specific Instruction.
```
mickeyspiegel: If we add in an "MD Length", some text about the effect of Domain Specific Metadata on MD Length belongs here.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  Domain Specific ID. Domain Specific Instruction is an instruction that requires additional 
+  processing of Domain Specific Flags (DS Flags) for the INT Domain identified by Domain Specific ID. 
+  If the Domain Specific ID matches any Domain ID known to this node, then additional processing
+  of the Domain Specific Flags and Domain Specific Instruction is required and Domain Specific
+  Metadata is also sent to the monitoring systems.
+
+  Some metadata can be defined as 'source-only' metadata by the INT domain. Those metadata
+  will be sent only by the INT source, not by INT transit or sink devices.
+  In a sense, 'source-only' bits do not serve as instructions for downstream INT devices
+  to follow. In INT-MX mode, the source node should not set the bits corresponding to
```
mickeyspiegel: s/should/must

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+  processing of Domain Specific Flags (DS Flags) for the INT Domain identified by Domain Specific ID. 
+  If the Domain Specific ID matches any Domain ID known to this node, then additional processing
+  of the Domain Specific Flags and Domain Specific Instruction is required and Domain Specific
+  Metadata is also sent to the monitoring systems.
+
+  Some metadata can be defined as 'source-only' metadata by the INT domain. Those metadata
+  will be sent only by the INT source, not by INT transit or sink devices.
+  In a sense, 'source-only' bits do not serve as instructions for downstream INT devices
+  to follow. In INT-MX mode, the source node should not set the bits corresponding to 
+  "source-only" metadata in the DS Instruction bitmap. However, the source node should ensure
```
mickeyspiegel: s/should/must
...
when such "source-only" metadata is included.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+
+  The amount of Domain Specific Metadata sent by each hop must be a multiple of 4 bytes,
+  determined from the Domain Specific Instruction. 
+  
+  If the Domain Specific ID does not match any Domain ID known to this node, then 
+  the transit node is required to either:
+  
+  - Send the metadata corresponding to INT Baseline Instruction, or
+      
+  - Skip INT processing altogether and not send any metadata the monitoring systems.
```
mickeyspiegel, line 1379: This does not match the text in Telemetry Report 2.0. I see three options, assuming that "MD Length" is included in the INT-MX header:
- Send the metadata corresponding to INT Baseline Instruction, ignoring the DS Instruction and setting the corresponding bits in the Telemetry Report header to zero,
- Send the metadata corresponding to INT Baseline Instruction, and pad with the special all-ones reserved value in lieu of DS Metadata, out to MD Length, or
- Skip INT processing altogether and not send any metadata to the monitoring systems.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+
+* Each INT Transit device along the path that supports INT-MX sends its own metadata
+values as specified in the instruction bitmap and DS instructions.
```
mickeyspiegel: The procedures below assume no "MD Length". They certainly should be described as one option. The other option would be based on "MD Length" and padding. IMO implementations should be allowed to choose either behavior.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+    are not set in the Telemetry Report sent to the monitoring system.
+  
+    Details of the metadata semantics and format of the Telemetry Report can 
+    be accessed at the link below:
+    
+    https://github.com/p4lang/p4-applications/blob/master/docs/telemetry_report_v2.0.pdf
+    
+* Summary of the field usage
+  - The INT Source must set the following fields:
+    - Ver, D, M, and Instruction Bitmap.
```
mickeyspiegel: M is always zero, I would not include it in this list.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+    - INT Source must set all reserved bits to zero.
```
mickeyspiegel: Here it says "must", up above in the definition of R it says "should". We need to pick one and be consistent.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1251,128 @@ Metadata if added by the source. The total stack length can be determined
...
+    - INT Source may set the Domain-specific fields. 
+  - Intermediate transit devices should not modify the INT-MX header.
```
mickeyspiegel: s/should/must

If you want the ability for DS Flags to be modified at transit, then you need to call that out. Domain Specific ID and DS Instruction must not be modified at transit.

telemetry/specs/INT.mdk
```diff
@@ -1330,6 +1483,31 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
    |                          TCP payload                          |
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
```
mickeyspiegel, line 1541: IMO it would be cleaner to add a section header. Change the title of the previous example to "Example with INT-MD over TCP", then add a title here for "Example with INT-MX over TCP", referring to the previous section for the IP and TCP headers. Start with the INT shim header with type = 3 and length = 3 here.

telemetry/specs/INT.mdk
```diff
@@ -1330,6 +1483,31 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
...
+INT-MX Metadata Header, followed by TCP payload:
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   | Ver=2 |0|0|0|       Reserved        | HopML=2 |RemainingHopC=6|
```
mickeyspiegel: RemainingHopC should not exist. For HopML / MD Length, see comments above.

telemetry/specs/INT.mdk
```diff
@@ -1330,6 +1483,31 @@ INT Metadata Header and Metadata Stack, followed by TCP payload:
...
+## Example with new UDP header and INT inserted before TCP
+
+As before we consider a scenario where host1 sends a TCP packet to host2.
+The ToR switch
+of host1 (Switch1) acts as the INT source. It adds a new UDP header,
+INT headers and its own metadata in the packet.
+Switch2 prepends its metadata. Finally, the ToR
+switch of host2 (Switch3) acts as the INT sink and removes the UDP and
+INT headers before forwarding the packet to host2.
+
```
mickeyspiegel: Somehow a redundant heading and first paragraph were inserted. They need to be deleted.

telemetry/specs/INT.mdk
```diff
@@ -1407,6 +1585,21 @@ INT Metadata Header and Metadata Stack, followed by TCP header:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
+INT-MX Metadata Header, followed by TCP header:
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   | Ver=2 |0|0|0|       Reserved        | HopML=2 |RemainingHopC=6|
```
mickeyspiegel: RemainingHopC should not exist. For HopML / MD Length, see comments above.

telemetry/specs/INT.mdk
```diff
@@ -1483,6 +1677,22 @@ INT Metadata Header and Metadata Stack, followed by UDP payload:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
+
```
mickeyspiegel, line 1846: IMO it would be cleaner to add a section header. Change the title of the previous example, replacing "INT" with "INT-MD", then add a title here with "INT-MX", referring to the previous section for the IP and UDP headers. Start with the INT shim header with type = 3 and length = 3 here.

telemetry/specs/INT.mdk
```diff
@@ -1483,6 +1677,22 @@ INT Metadata Header and Metadata Stack, followed by UDP payload:
...
+INT-MX Metadata Header, followed by UDP payload:
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   | Ver=2 |0|0|0|       Reserved        | HopML=2 |RemainingHopC=6|
```
mickeyspiegel: RemainingHopC should not exist. For HopML / MD Length, see comments above.

telemetry/specs/INT.mdk
```diff
@@ -1538,6 +1750,42 @@ the GRE was inserted by the INT source.
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
mickeyspiegel, line 1957: The example for INT-MD over IPv4/GRE is missing the first 4 bytes of the INT-MD header, after the shim header.

telemetry/specs/INT.mdk
```diff
@@ -1538,6 +1750,42 @@ the GRE was inserted by the INT source.
...
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  |
+   |     Checksum (optional)       |        Offset (Optional)      |  |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  G
+   |                         Key (Optional)                        |  R
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  E
+   |                   Sequence Number (Optional)                  |  |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+  |
+   |                        Routing (Optional)                     |  |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+   |Type=1 |1| Rsvd|    Length=7   |       Protocol = 0x800        |  |
```
mickeyspiegel: Type=3 and length = 3

The first 4 bytes of the INT-MX header missing.

telemetry/specs/INT.mdk
```diff
@@ -1407,6 +1585,21 @@ INT Metadata Header and Metadata Stack, followed by TCP header:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
```
mickeyspiegel, line 1690: IMO it would be cleaner to add a section header. Change the title of the previous example, replacing "INT" with "INT-MD", then add a title here with "INT-MX", referring to the previous section for the IP and TCP headers. Start with the INT shim header with type = 3 and length = 3 here.

telemetry/specs/INT.mdk
```diff
@@ -1538,6 +1750,42 @@ the GRE was inserted by the INT source.
...
 
+
```
mickeyspiegel, line 1972: IMO it would be cleaner to add a section header. Change the title of the previous example, replacing "INT" with "INT-MD", then add a title here with "INT-MX".

telemetry/specs/INT.mdk
```diff
@@ -1596,6 +1844,18 @@ INT Metadata Header and Metadata Stack:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
```
mickeyspiegel, line 2067: IMO it would be cleaner to add a section header. Change the title of the previous example, replacing "INT" with "INT-MD", then add a title here with "INT-MX", referring to the previous section for the VXLAN-GPE header. Start with the INT shim header with type = 3 and length = 3 here.

telemetry/specs/INT.mdk
```diff
@@ -1596,6 +1844,18 @@ INT Metadata Header and Metadata Stack:
...
 
+INT-MX Metadata Header:
+`
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   | Ver=2 |0|0|0|       Reserved        | HopML=2 |RemainingHopC=5|
```
mickeyspiegel: RemainingHopC should not exist. For HopML / MD Length, see comments above.

mickeyspiegel (changes requested):
Mostly minor things, but also some technical comments, primarily around "Hop ML" / "MD Length" and padding options.

Some other cleanup that should be done before tagging INT 2.1, not directly related to this PR:
- s/Switch id/Node id
- References to INT-CLONE-MD and INT-PROBE-MD need to be deleted since these terms are no longer defined in the INT spec

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Changed.

telemetry/specs/INT.mdk
rsivakolundu: Corrected

telemetry/specs/INT.mdk
rsivakolundu: Changed.

telemetry/specs/INT.mdk
rsivakolundu: I agree with that change. Keeping it generic allows for probes.

telemetry/specs/INT.mdk
rsivakolundu: Help for future revisions to not reuse the bits.

telemetry/specs/INT.mdk
rsivakolundu: Help for future revisions to not reuse the bits.

telemetry/specs/INT.mdk
rsivakolundu: What if someone implements only INT-MX. This would mean that Source Node has to compute the Metadata requested and program the Hop_ML field. For INT-MX, the node generating the Telemetry Report has to gather the metadata requested and compute the length anyways. So setting the MD Length field in the Telemetry Report shouldn't be an issue. Also, what if the metadata requested is not available at a node, then that node will have to compute the M<D Length accordingly anyways. Hop_ML is critical in INT-MD, not in INT-MX. Thoughts?

telemetry/specs/INT.mdk
rsivakolundu: Not following your comment.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: See my comment about Hop_ML in INT-MX.

telemetry/specs/INT.mdk
rsivakolundu: Deleted.

telemetry/specs/INT.mdk
rsivakolundu: Agree. Fixed.

telemetry/specs/INT.mdk
rsivakolundu: Added text.

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
rsivakolundu: Corrected.

telemetry/specs/INT.mdk
rsivakolundu: Corrected

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
mickeyspiegel: I am proposing the addition of one requirement at the Source Node, to set MD_Length, and then allowing a variety of methods at transit and sink nodes. If you don't want to look at MD_Length in the INT-MX header at transit and sink, you don't have to.

I am proposing two different implementation options at transit and sink nodes implementing INT-MX:

1. As you mention, when requested metadata is not available or one or more bits are not understood, bits could be cleared when copying into the Telemetry Report's RepMdBits and DSMdBits. In that case MD_Length in the Telemetry Report would need to be recomputed. MD_Length was put into the Telemetry Report to ease in parsing at the collector, in case the collector does not know or understand some bit definitions in DSMdBits, so that the collector can find the start of the inner report contents such as truncated packet fragments. This is purposely meant to be similar to the way Hop_ML in INT-MD is used by the collector to distinguish metadata from different hops.

2. An alternative would be to do what INT-MD does, which is to leave all the bits set, and insert the special all-ones value for unavailable or unsupported metadata. This is why MD_Length must be inserted by the source node, so that these implementations can determine how much padding to add. With this approach, there is no need to recompute MD_Length.

I am not sure I see a reason to preclude a hybrid, where some bits are cleared and others have the special all-ones value, since that would still be parseable at the collector. However, I am not sure why one would do that.

telemetry/specs/INT.mdk
```diff
@@ -1093,7 +1093,8 @@ INT-MD Metadata Header and Metadata Stack:
 
 * The fields in the INT metadata header are interpreted the following way:
   - Ver (4b): INT metadata header version. Should be 2 for this version.
-  - D   (1b): Discard Copy/Clone. INT Sink should Discard the packet after Extracting INT data 
+  - D (1b): Discard. 
+    - INT Sink should Discard the packet after Extracting INT-MD metadata.
```
mickeyspiegel: s/should/must

telemetry/specs/INT.mdk
```diff
@@ -1364,10 +1368,12 @@ values as specified in the instruction bitmap and DS instructions.
...
 * Summary of the field usage
   - The INT Source must set the following fields:
-    - Ver, D, M, and Instruction Bitmap.
-    - INT Source must set all reserved bits to zero.
+    - Ver, D and Instruction Bitmap.
+    - INT Source should set all reserved bits to zero.
     - INT Source may set the Domain-specific fields. 
-  - Intermediate transit devices should not modify the INT-MX header.
+  - Intermediate transit devices may set bits in the DS Flags field. Intermediate 
+    transit devices should not modify the Domain Specific ID and
```
mickeyspiegel: s/should/must

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: I think I corrected it.

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
```diff
@@ -1287,9 +1287,29 @@ be accessed at the link below:
...
+    Domain Specific Metadata in 4-Byte words to be sent to monitoring systems at 
+    each INT transit hop.
+    - Hop ML is set by the INT source. Transit hops and INT sink node may refer to it.
+      If an INT domain uses 'source-only' Domain Specific Metadata,
+      defined below, the length of the source-only Domain Specific Metadata
+      is *excluded* from the Hop ML.
+    - The largest value of Hop ML for baseline and domain specific
+      metadata is 31. An INT-capable device may be limited in the maximum number 
+      of instructions it can process and/or maximum length of metadata it can 
+      gather for each data packets. An INT hop that cannot process all instructions '
```
mickeyspiegel, line 1309: nit: Remove extraneous ' at the end of the line.

telemetry/specs/INT.mdk
```diff
@@ -1287,9 +1287,29 @@ be accessed at the link below:
...
+      have the following alternatives:
```
mickeyspiegel, line 1310: nit: s/have/has

telemetry/specs/INT.mdk
mickeyspiegel: Line 484
"The format of this header will be defined in the next minor revision."
Replace with
"The format of this header is defined in section [#sec-int-mx-metadata-header-format]."

telemetry/specs/INT.mdk
```diff
@@ -484,20 +483,52 @@ are present, the MD/MX-type header must precede the Destination-type header.
...
+In the INT-MD mode, each device in the packet forwarding path creates additional 
+space in the INT-MD Header on-demand to add its own INT metadata. To avoid 
+exhausting header space in the case of a forwarding loop or any other anomalies, 
+each INT Transit Hop must decrement the Remaining Hop Count field in the INT 
+header appropriately.
+
+In the INT-MX mode, each device in the packet forwarding path follows the 
+intructions in the INT-MX Header, gathers the device specific metadata and 
+exports the device metadata using the Telemetry Report.  INT Transit Hop 
+nodes may update the DS Flags field in the INT-MX header. Domian Specific ID
```
mickeyspiegel, line 513: typo: s/Domian/Domain

telemetry/specs/INT.mdk
mickeyspiegel: Starting at line 561 in the MTU section, there is a paragraph that currently reads:

"Regardless of whether or not an INT transit switch participates in Path MTU
discovery, if it cannot insert all requested metadata because doing so will
cause the packet length to exceed egress link MTU, it
must not insert any metadata and set the M bit in the
INT header, indicating that egress MTU was exceeded at an INT hop."

This could be changed to something like:

"Regardless of whether or not an INT transit switch participates in Path MTU
discovery, if it cannot insert all requested metadata because doing so will
cause the packet length to exceed egress link MTU, it must either:
- not insert any metadata and set the M bit in the INT header, indicating
  that egress MTU was exceeded at an INT hop, or
- report the metadata collected from previous hops (setting the
  Intermediate Report bit if a Telemetry Report 2.0 packet is generated)
  and remove the reported metadata from the packet, including the
  metadata from this transit hop in either the report or the INT-MD
  metadata header."

telemetry/specs/INT.mdk
mickeyspiegel: ... determined from the Domain Specific Instruction or consistent with the per-hop
metadata length (Hop ML) set by the INT source (see Hop ML above).

telemetry/specs/INT.mdk
mickeyspiegel: Not changed yet.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1252,152 @@ Metadata if added by the source. The total stack length can be determined
...
+be accessed at the link below:
+  https://github.com/p4lang/p4-applications/blob/master/docs/telemetry_report_v2.0.pdf
+
+* The fields in the INT metadata header are interpreted the following way:
+  - Ver (4b): INT metadata header version. Should be 2 for this version.
+  - D (1b): Discard. 
+    - INT Sink must Discard the packet after sending the metadata requested
+      in the INT-MX header to the monitoring system.
+  - E (1b): Max Hop Count exceeded. 
+    - Not applicable in INT-MX mode. Should be set to 0 by the INT Source.
```
mickeyspiegel: s/should/must

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1252,152 @@ Metadata if added by the source. The total stack length can be determined
...
+  - M (1b): MTU exceeded. 
+    - Not applicable in INT-MX mode. Should be set to 0 by the INT Source.
```
mickeyspiegel: s/should/must

telemetry/specs/INT.mdk
mickeyspiegel: I would make the text starting with "An INT hop that cannot process ..." a new bullet

telemetry/specs/INT.mdk
mickeyspiegel: While the "Hop ML" text above was modified, this text has not been modified. I suggest the text that I proposed above in this comment.

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1252,152 @@ Metadata if added by the source. The total stack length can be determined
...
+  the transit node is required to either:
+  
+  - Send the metadata corresponding to INT Baseline Instruction, or
+      
+  - Skip INT processing altogether and not send any metadata the monitoring systems.
+
+* Each INT Transit device along the path that supports INT-MX sends its own metadata
+values as specified in the instruction bitmap and DS instructions.
+  - If a device is unable to provide a metadata value specified in the
+    instruction bitmap because its value is not available, it must ensure that the
```
mickeyspiegel: it must either:
- ensure that the corresponding bit in "RepMdbits" and/or "DSMdbits" is not set in the
  Telemetry Report, or
- send the all-ones reserved value (4 or 8 bytes of 0xFF depending on the length of metadata).

telemetry/specs/INT.mdk
```diff
@@ -1220,6 +1252,152 @@ Metadata if added by the source. The total stack length can be determined
...
+    corresponding bit in the "RepMdbit" and/or "DSMdbit" is not set in the Telemetry Report.
+  
+    Details of the metadata semantics and format of the Telemetry Report can 
+    be accessed at the link below:
+    
+    https://github.com/p4lang/p4-applications/blob/master/docs/telemetry_report_v2.0.pdf
+    
+  - Reserved bits in the instruction bitmap are to be handled as described. If an
+    INT transit hop receives a reserved bit set in the instruction bitmap (e.g.
+    set by a INT source that is running a newer version), the transit hop must
```
mickeyspiegel: the transit hop must either:
- ensure that the corresponding reserved bits in "RepMdbits" and/or "DSMdbits" 
  are not set in the Telemetry Report sent to the monitoring system, or
- for "RepMdBits", send the all-ones reserved value (4 bytes of 0xFF);
  for "DSMdbits", pad with the special all-ones reserved value in lieu of reserved
  DS Metadata, out to MD Length.

telemetry/specs/INT.mdk
```diff
@@ -1407,14 +1657,80 @@ INT Metadata Header and Metadata Stack, followed by TCP header:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
-## Example with INT in-between UDP header and UDP payload
+## Example with new UDP header and INT-MX inserted before TCP
+
+As before we consider a scenario where host1 sends a TCP packet to host2.
+The ToR switch of host1 (Switch1) acts as the INT source. It adds a new UDP header,
```
mickeyspiegel: nit: s/UDP header, INT-MX header/UDP header and a new INT-MX header

telemetry/specs/INT.mdk
```diff
@@ -1538,15 +1926,65 @@ the GRE was inserted by the INT source.
...
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<-+
+   |      Inner IPv4 Header + Payload + Padding (L2/L3/ESP/….)     |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+`
+
+## Example with INT-MD over VXLAN GPE
 
 We now consider a scenario where Host1 and Host2 use VXLAN encapsulation.
-Host1 acts as VXLAN tunnel endpoint and INT source, inserts VXLAN and INT
+Host1 acts as VXLAN tunnel endpoint and INT source, inserts VXLAN and INT-MX
```
mickeyspiegel: s/INT-MX/INT-MD

telemetry/specs/INT.mdk
```diff
@@ -1538,15 +1926,65 @@ the GRE was inserted by the INT source.
...
 headers with instruction bits corresponding to the network state to be
 reported at intermediate switches. In this example, Host1 itself does
 not insert any INT metadata. Intermediate switches parse through VXLAN
 header and populate the INT metadata. Host2 acts as INT sink and VXLAN
-tunnel endpoint, removes INT and VXLAN headers.
+tunnel endpoint, removes INT-MX and VXLAN headers.
```
mickeyspiegel: s/INT-MX/INT-MD

telemetry/specs/INT.mdk
```diff
@@ -1596,15 +2034,62 @@ INT Metadata Header and Metadata Stack:
...
 
-## Example with INT over Geneve
+## Example with INT-MX over VXLAN GPE
+
+We now consider a scenario where Host1 and Host2 use VXLAN encapsulation.
+Host1 acts as VXLAN tunnel endpoint and INT source, inserts VXLAN and INT-MX
+header with instruction bits corresponding to the network state to be
+reported at intermediate switches. In this example, Host1 itself does
+not send any metadata to the monitoring system. Intermediate switches parse 
+through VXLAN header and and processes the INT-MX header and sends the INT
```
mickeyspiegel: s/and and processes/, process
s/sends/send

telemetry/specs/INT.mdk
```diff
@@ -1655,6 +2140,54 @@ INT Metadata Header and Metadata Stack:
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 `
 
+## Example with INT-MX over Geneve
+
+Finally, we consider a scenario where Host1 and Host2 use Geneve encapsulation.
+Host1 acts as Geneve tunnel endpoint and INT source, inserts Geneve and INT-MX
+header with instruction bits corresponding to the network state to be
```
mickeyspiegel: nit: s/header/headers

telemetry/specs/INT.mdk
```diff
@@ -1655,6 +2140,54 @@ INT Metadata Header and Metadata Stack:
...
+reported at intermediate switches. In this example, Host1 does not send INT
+metadata to the monitoring system. Intermediate switches parse through Geneve
+header and processes the INT-MX header and sends the INT metadata requested
```
mickeyspiegel: s/and processes/, process
s/sends/send

telemetry/specs/INT.mdk
```diff
@@ -449,17 +449,16 @@ https://github.com/p4lang/p4-applications/blob/master/telemetry/code/models/p4-d
 # INT Headers
 
 This section specifies the format and location of INT Headers.
-INT Headers and their locations are relevant for
-INT-MX, INT-MD, INT-CLONE-MD, and INT-PROBE-MD modes
-where the INT instructions (and metadata stack in case of MD modes)
-are written into the packets.
+INT Headers and their locations are relevant for INT-MX, INT-MD modes where
```
mickeyspiegel: nit: s/INT-MX, INT-MD/INT-MX and INT-MD

mickeyspiegel (changes requested):

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Changed

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk
rsivakolundu: Added.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: Changed

telemetry/specs/INT.mdk
rsivakolundu: Done

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
rsivakolundu: I think I got it  correct now.

telemetry/specs/INT.mdk
rsivakolundu: Done now.

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
rsivakolundu: Changed.

mickeyspiegel (approved):

telemetry/specs/INT.mdk
rsivakolundu: Fixed

telemetry/specs/INT.mdk
rsivakolundu: Done.

telemetry/specs/INT.mdk
rsivakolundu: Changed

PR #81, opened 2020-05-15
Telemetry Report v2.0 final version

mickeyspiegel:
- Separated the Telemetry Report Header description into separate sections
  for 'Telemetry Reports Outer Header' (first 8 octets that appear once in
  a packet, and 'Individual Report Header' that may be repeated, one per
  report in the packet.
- Within the 'Individual Report Header', created subsections for
  'Telemetry Report Contents' when 'RepType' == 'INT', and for
  'Inner Report Contents' when 'InType' == 'TLV'
- Reworded section on Embedded Telemetry Metadata to match the v2.0
  structure, where the truncated packet fragments reside within each
  Individual Report Header
- Lots of minor editorial and formatting cleanup
- Added definition of 'node', and changed 'switch' and 'device' to
  'node' where appropriate
- Changed 'aggregated' vs 'individual' terminology to 'stacked' vs 'per hop'
- Reserved Domain Specific ID value 0x0000 as default well known ID,
  for which all DSMdBits are treated as reserved
- Updated references
- Added telemetry_report_latest.pdf that links to telemetry_report_v2_0.pdf


mickeyspiegel:
There are still some things left to do, but the sections reorganizing the descriptions of the telemetry report header format are ready and should be reviewed.

Some open items:
- The Parsing Considerations section does not align with the current format, either needs to be revised or dropped.
- The examples are pretty rough, should probably go to the same level of detail as some of the INT examples with respect to metadata, but transport can be ignored.
- The whole spec could use a re-read and scrub for editorial fixes and any lingering misalignment due to text from older versions

mickeyspiegel:
Also need a new INT-MX figure

rsivakolundu (commented):
Looks good to me. Thanks for cleaning it up.

telemetry/specs/telemetry_report.mdk
```diff
@@ -439,54 +456,78 @@ The Destination Port is user configurable. The expectation is that the same
...
     intended destination.
 
 * Node ID (32b): Node ID
-  : The unique ID of a network device (such as routers, switches, NICs etc.) This is generally administratively assigned. Node IDs
-    must be unique within a management domain.
+  : The unique ID of a node. This is generally administratively assigned.
+    Node IDs must be unique within a management domain.
+
+## Individual Report Header (Ver 2.0) (4+ octets) { #individual-report }
+
```
jklr, line 533: Can we spell out the rationale behind this report coalescing? This is to reduce the # of report pkts in the network etc. IMO, "coalescing" is a keyword but it's mentioned only once in the change log at the end of this spec.

telemetry/specs/telemetry_report.mdk
```diff
@@ -439,54 +456,78 @@ The Destination Port is user configurable. The expectation is that the same
...
+  |                                                               | Length
+  |                     Inner Report Contents                     |   |
+  |       (Truncated Packet or Additional DS Extension Data       |   |
+  |                  or TLV depending on InType)                  |   |
+  |                                                               |   |
+  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+<--+
+```
 
 * RepType (4b): Report Type
   : Type of Telemetry report
```
jklr: To clarify, how about
"Type of each individual report"
?

telemetry/specs/telemetry_report.mdk
```diff
@@ -439,54 +456,78 @@ The Destination Port is user configurable. The expectation is that the same
...
-    * 1: IOAM
-    * 2 – 15: Reserved
+    * **0**: Inner Only
+    * **1**: INT
+    * **2**: IOAM
+    * **3 – 15**: Reserved
 
 * InType (4b): Inner Type
-  : Type of data embedded after Variable Optional Baseline & DS Metadata.
+  : Type of data embedded after Telemetry Report Contents:
```
jklr: how about "... Individual Report Contents"

telemetry/specs/telemetry_report.mdk
```diff
@@ -496,41 +537,103 @@ particular deployment.
     * 6-15: Reserved.
 
 * Report Length (8b): Report Length
-  : Indicates length of report header in multiple of 4-byte words, including the variable 
-    optional baseline metadata,  DS metadata and Truncated Packet or additional Domain 
-    Specific Extension Data, and excluding the 4-byte word (RepType, InType, Report Length, 
-    MD Length, D, Q, F, I, Rsvd).
+  : Indicates the length of the Individual Report Header in a
+    multiple of 4-byte words, including the *Telemetry Report Contents* and
```
jklr: Individual Report Contents

telemetry/specs/telemetry_report.mdk
jklr: I see there are multiple occurrences of this term below. Would need to replace all.

telemetry/specs/telemetry_report.mdk
```diff
@@ -550,158 +653,178 @@ particular deployment.
...
-0 – Domain Specific Extension Data
-1 - Ethernet
-2 – IPv4
-3 – IPv6
-4-15 – Reserved.
+* Variable Optional Domain Specific Metadata
+  : The metadata corresponding to the *DSMdBits*, 4 octets or a multiple of
+    4 octets for each bit.
+
+### Inner Report Contents for *InType* 1 (*TLV*) (4+ octets) { #tlv }
```
jklr: Inner Report Contents: 
Another term used to indicate "Telemetry Report Contents" or "Individual Report Contents". Can we unify them?

telemetry/specs/telemetry_report.mdk
```diff
@@ -550,158 +653,178 @@ particular deployment.
...
+    indicated by *TLVType* codepoint 1, 2, or 3, which identifies the type of
+    header at the beginning of the truncated packet.
 
-TLVLength (8b) - Indicates the length, in 4-byte word, of Variable TLV Data as identified by TLVType.
+* Domain Specific Extension Data
+  : Domain Specific Extension Data, whose format can be determined from the
+    *Domain Specific ID* specified in the *Telemetry Report Contents* and the
+    *TLV Data Template*. For *RepType* codepoint 1 *INT*, this is additional
+    domain specific data, not associated with *DSMdBits*.
+    Presence of this field is indicated by *TLVType* codepoint 0.
```
jklr, line 807: These different combinatory cases are quite confusing.
Wonder if we can add a table to summarize them?

telemetry/specs/telemetry_report.mdk
mickeyspiegel: "Inner Report Contents" is *not* the same as "Telemetry Report Contents".

How about:
"Telemetry Report Contents" => "Individual Report Primary Contents" or "Individual Report Main Contents" or "Individual Report Principal Contents"
"Inner Report Contents" => "Individual Report Inner Contents"

jklr: that's right.

I like this:
"Telemetry Report Contents" => "Individual Report Main Contents"
"Inner Report Contents" => "Individual Report Inner Contents"

telemetry/specs/telemetry_report.mdk
```diff
@@ -819,18 +943,39 @@ specification.
...
-  - Added changes for v2.0
+  - Revised Telemetry Report header 2.0 format supporting:
+    - coalescing of multiple reports in one packet
+    - support for domain specific extensions
+    - some new RepMdBits codepoints
+    - Increased timestamp size to 8 bytes
+  - Changed *Switch id* terminology to *Node ID*
+* 2020-05-15
+  - Separated the Telemetry Report Header description into separate sections
+    for 'Telemetry Reports Outer Header' (first 8 octets that appear once in
```
jklr: The use of outer is

telemetry/specs/telemetry_report.mdk
jklr: As a 2nd thought, the term "inner" seems confusing because we have "Telemetry Reports Outer Header" up there. The outer header is for the entire coalesced report pkt, and the inner header is for each individual report. The outer vs inner are applied to two different objects.

One idea is to change
"Telemetry Reports Outer Header" => "Telemetry Report Group Header", 
where a group may have one or more individual reports in it. 
We can define the term 'group' first and use it else where.

rsivakolundu:
I like the Group Header wording.

-Ramesh

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Added a new subsection "Coalescing A Group of Telemetry Reports In A Single Packet", at the end of "Key Concepts".

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Done

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Done

mickeyspiegel:
From my perspective, all that is left now is the examples, INT-MX figure, and resolution of any new comments from this point.

rsivakolundu:
How can I help?

-Ramesh

mickeyspiegel:
@rsivakolundu Put some details in your examples, along the lines of what is in INT 2.1.

I cannot put any time into this tomorrow, not yet sure about Wednesday.

rsivakolundu:
Will spend some time on it tomorrow.

Thanks

-Ramesh

jklr (approved):
The changes LGTM.

mickeyspiegel:
Done adding examples and figures. This is ready for final review.

telemetry/specs/telemetry_report.mdk
```diff
@@ -398,9 +402,53 @@ corresponding bits are set.
...
+truncated packet fragment all the way down past any embedded telemetry
+metadata (if present), even when the Individual Report includes optional
+metadata such as drop reason. It may also want to process the embedded
+telemetry metadata, for example to recognize the case where a path change
+directs traffic to a congested node where packets are being dropped.
+
+## Coalescing A Group of Telemetry Reports In A Single Packet
+
+Starting with Version 2.0, the telemetry report format allows for a group
+of individual reports, each corresponding to one data plane packet, to be
```
rsivakolundu: Can we reword it as 

of individual reports, each corresponding to one flow (multiple packets) instead of one data plane packet.

telemetry/specs/telemetry_report.mdk
```diff
@@ -496,41 +577,103 @@ particular deployment.
...
+* Rsvd (4b): Reserved
   : Should be set to zero upon transmission, and ignored upon reception
 
+* Individual Report Main Contents
+  : The metadata that comprises this report, along with associated fields
+    that assist in processing the metadata. The format varies depending on
+    *RepType*.
+
+    When the *RepType* value is *Inner Only*, then the Individual Report Main
+    Contents is empty. *MD Length* should be set to zero upon transmission,
```
rsivakolundu, line 628: When MD Length is zero, even the RepMdBits and Domain Specific ID, DSMdBits, DSMdStatus fields are also not there. However, the MD Length is defined as the length of the Variable Optional Baseline Metadata and DS Metadata. The 2 quad-byte words corresponding to RepMdBits and Domain Specific ID, DSMdBits, DSMdStatus  are not included in the MD Length. We need to capture this somewhere.

telemetry/specs/telemetry_report.mdk
mickeyspiegel: I don't think I agree to that.

How about "one data plane packet or one flow"?

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Your first sentence is not true. Presence of RepMdBits, Domain Specific ID, DsMdBits, DsMdstatus is based on RepType value of INT.

telemetry/specs/telemetry_report.mdk
rsivakolundu: I am ok with "one data plane packet or one flow".

telemetry/specs/telemetry_report.mdk
rsivakolundu: I see. Individual Report Type Contents for RepType value of INT defines those fields. If the RepType value is anything else, then the format is different. Makes sense.

rsivakolundu (approved):

mickeyspiegel:
@rsivakolundu We should be clear about TLVLength. Does it include the first four bytes starting with TLVType and TLVLength or not?

rsivakolundu (approved):

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Done

telemetry/specs/telemetry_report.mdk
mickeyspiegel: Added a sentence clarifying that the length does not include the fixed portion of Individual Report Main Contents.

PR #83, opened 2020-06-10
INT-MX 'source-only' metadata embedded by source and reported by each node

mickeyspiegel:
* In the INT-MX header, unlike other metadata, 'source-only' domain
  specific metadata is embedded in the packet after "DS Instruction"
  and "DS Flags". This 'source-only' metadata should be reported by
  each node including the source, transit, and sink nodes.

* In the Telemetry Report, the 'source-only' metadata may be reported
  in a truncated packet containing the INT-MX header, in domain specific
  extension data, or in "Variable Optional Domain Specific Metadata"
  with the corresponding bit set in "DSMdbits".

* In each of the "INT-MD Metadata Header Format" and "INT-MX Header
  Format" sections, consolidated the transit node processing text:
  - Moved text on INT-capable node limited in instructions or inserted
    metadata length, from Hop ML to transit node processing bullet.
  - Moved text on unknown Domain Specific ID from DS Instruction to
    transit node processing bullet.
  - In INT-MD, consolidated the text on INT-capable node limited in
    instructions or inserted metadata length with the text on a node
    that cannot process all required metadata, since they are redundant.
  - In INT-MD, clarified the order of metadata insertion in the
    transit node processing bullet.

* Reorganized INT-MX transit node processing text by splitting 2 options:
  1. Modify "RepMdBits" and "DsMdbits" in the Telemetry Report, after
     copying from INT-MX "Instruction Bitmap" and "DS Instruction"
     respectively, and recompute "MD Length".
  2. Copy INT-MX "Instruction Bitmap" and "DS Instruction" to
     "RepMdBits" and "DSMdbits" in the Telemetry Report, make minimal
     modifications as required, and adjust "MD Length".
  Made these two options the high level bullets, and then the detailed
  processing for unavailable metadata, reserved metadata, unknown
  "Domain Specific ID", and max exceeded become the inner bullets under
  each of the two options.


mickeyspiegel:
The first two commits are reorganization of existing text.
The third commit contains the technical changes.

mickeyspiegel:
There are lots of tricky details regarding unknown Domain Specific ID and backwards compatibility implications. Pay attention when reviewing.

mickeyspiegel:
After splitting the two main options for INT-MX transit node processing, there is a question whether the second option with padding is too complicated.
Should it be deleted?

jklr:
Thanks Mickey. I will review.
I vote for deleting the second option with padding, for the sake of simplicity.

rsivakolundu (approved):

rsivakolundu (commented):
I also vote for removing option #2.

jklr:
If we remove option #2, we can remove HopML from the MX header.
 
Q1. Can we also remove E and M bits and make them all reserved?
Q2. how about adding 24b SEQ# in the freed-up space? SEQ# seems commonly needed by edge-to-edge and also some congestion control use cases.

telemetry/specs/INT.mdk
```diff
@@ -1342,6 +1339,9 @@ INT-MX Header:
 [^telem-report]: [Telemetry Report Format Specification Version 2.0](https://github.com/p4lang/p4-applications/blob/master/docs/telemetry_report_v2_0.pdf), May 2020.
 
 * **Ver (4b):** INT-MX header version. Should be 2 for this version.
+
+[]{tex-cmd: "\newpage"}
+
```
jklr, line 1344: "Bits 0 - 14 are Baseline INT Instructions." at the end of Instruction Bitmap seems not needed. I presume the quoted text was added for INT-MD which has Bit15, checksum complement.

mickeyspiegel:
@jklr We can remove HopML if it is no longer used when forming Telemetry Reports.

I believe @rsivakolundu wanted to keep the E and M bit definitions when I asked a while ago.

I don't believe seq # should always be there. It makes sense from NICs, but it is difficult for switches to insert. With all the changes to allow source-only metadata, that seems like a perfect place for seq # to go.

telemetry/specs/INT.mdk
mickeyspiegel: It is baseline metadata compared to the domain specific metadata. In the Telemetry Report with RepType INT, the baseline metadata goes into "Variable Optional Baseline Metadata" and the domain specific metadata goes into "Variable Optional Domain Specific Metadata". I prefer keeping it.

rsivakolundu:
I am ok with removing E and M bit and HOP-ML field and making all of them reserved. Cannot deal with using those reserved fields for anything at this stage.

telemetry/specs/INT.mdk
jklr: I see. Then let's add your comment into the spec text. They help a lot.

jklr:
sure.
Let's remove E, M and HopML and keep them all as reserved bits.

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
-  to know how to parse and use the data.
+  Some metadata can be defined as 'source-only' metadata by the INT domain.
+  Unlike other metadata, 'source-only' metadata is embedded in the packet by the
+  INT source node, as *Optional Domain Specific 'Source-Only' Metadata*. The
+  *DS Instruction* bits corresponding to 'source-only' metadata do not serve as
+  instructions for downstream INT nodes to insert their own metadata. Instead,
+  they indicate that the metadata is embedded, and that each node including
+  the source, transit, and sink nodes should report that source metadata to the
+  monitoring system along with the node's own metadata. Note that there are
+  several ways to include the 'source-only' metadata in the Telemetry Report,
```
jklr, line 1410: "several ways" => pls update after removing the second option with padding.

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
 
   The amount of Domain Specific Metadata sent by each hop must be a multiple of 4 bytes,
   either determined from the *DS Instruction* or consistent with the
-  per-hop metadata length (*Hop ML*) set by the INT source (see *Hop ML* above).
-
-  []{tex-cmd: "\newpage"}
-
-  If the *Domain Specific ID* does not match any Domain ID known to this node, then
-  the transit node is required to either:
+  per-hop metadata length (*Hop ML*) set by the INT source, as described below.
```
jklr, line 1415: Hop ML will be dropped.

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
-      - ensure that the corresponding bit in *RepMdbits* and/or *DSMdbits* is not set
-        in the Telemetry Report [^telem-report], or
+  values as specified in the *Instruction Bitmap* and *DS Instruction*. Either
+  one of the following two procedures may be implemented:
+
+  1. Modify *RepMdBits* and *DSMdbits* in the Telemetry Report [^telem-report],
+     after copying from INT-MX *Instruction Bitmap* and *DS Instruction*
+     respectively, and recompute *MD Length*.
+
+     - If INT-MX *Instruction Bitmap* bit 0 was set, clear *RepMdBits* bit 0.
```
jklr: can we spell out the reason why? because Node ID will be part of the group header and is common, no need to repeat at every individual report. I believe we provide the context in the TR spec but repeating here won't hurt.

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
-
-  - If a node is unable to provide a metadata value specified in the
-    instruction bitmap because its value is not available, it must either:
-
-      - ensure that the corresponding bit in *RepMdbits* and/or *DSMdbits* is not set
-        in the Telemetry Report [^telem-report], or
+  values as specified in the *Instruction Bitmap* and *DS Instruction*. Either
+  one of the following two procedures may be implemented:
+
+  1. Modify *RepMdBits* and *DSMdbits* in the Telemetry Report [^telem-report],
```
jklr, line 1426: There are multiple layers/headers in TR 2.0. I guess RepMdBits and DSMdbits are terms used only for the individual main content header, hence there is no ambiguity.

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
+
+     - If the *Domain Specific ID* does not match any Domain ID known to this
+       node, then the transit node is required to either:
+       - Send the metadata corresponding to *Instruction Bitmap* and ensure that
+         *DSMdbits* is not set in the Telemetry Report and the *MD Length* field
+         reflects the length of the metadata sent in the Telemetry Report, or
+       - Skip INT processing altogether and not send any metadata to the
+         monitoring systems.
+
+     - The INT node should make every effort to include 'source-only' metadata
```
jklr, line 1448: "The INT node" => transit/sink only or does it also apply to source?
Asking because the next line says "that was received in the INT-MX header)".

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
+       (that was received in the INT-MX header) in the Telemetry Report. There
+       are several ways to accomplish this:
+       - The truncated packet in the *Individual Report Inner Contents* includes
+         the received INT-MX header with embedded 'source-only' metadata, or
+       - The domain specific extension data includes the 'source-only' metadata,
```
jklr: domain specific extension data
=> * Domain Specific Extension Data *

telemetry/specs/INT.mdk
```diff
@@ -1409,57 +1397,158 @@ INT-MX Header:
...
+       to be careful when leaving some bits in *DSMdbits* reserved, with regard
+       to defining any of those bits as 'source-only' in the future.
+
+     - An INT node may be limited in the maximum number of instructions it can
+       process and/or maximum length of metadata it can gather for each data
+       packet. An INT hop that cannot process all instructions must send to the
+       monitoring system the metadata it can process, updating the *RepMdBits*,
+       *DSMdBits* and *MD Length* fields appropriately in the Telemetry Report.
+
+  2. Copy INT-MX *Instruction Bitmap* and *DS Instruction* to *RepMdBits* and
```
jklr, line 1486: I guess this entire option 2 will be removed.

jklr (commented):
Can we modify the example (sec 6.5) to include DS SO metadata such as SEQ number? Thx!

telemetry/specs/INT.mdk
mickeyspiegel: The reason I did not do that is because there are two ways to report domain specific extension data, *Additional DS Extension Data* in *Individual Report Inner Contents*, or *Domain Specific Extension Data* in TLV in *Individual Report Inner Contents*. Similar thing for the truncated packet.

telemetry/specs/INT.mdk
mickeyspiegel: Done

telemetry/specs/INT.mdk
jklr: I see. It's a bit hard to make sense out of the free form "domain specific extension data" and the two different ways of doing it by just reading the TR spec.
 
@rsivakolundu 
do you need this 2nd option of using the "domain specific extension data" to carry MX source-only metadata? If not, I suggest we drop this option here to minimize confusion. 
No matter what we mention this option here or not, it's implicitly allowed anyway because the domain specific extension data is a complete free-form.

mickeyspiegel:
@jklr I threw an INT-MX with domain specific source-only metadata consisting of a Flow ID and Sequence Number into a separate example at the end. I also moved the INT-MD domain specific source-only example from Randy to the end of the Examples section instead of being stuck in an appendix.

telemetry/specs/INT.mdk

telemetry/specs/INT.mdk
mickeyspiegel: I don't see any ambiguity, will leave it as is.

telemetry/specs/INT.mdk
mickeyspiegel: The first level bullet is only transit, but this really applies to source, transit, and sink. Up in the DS Instruction definition, it does say:

> each node including the source, transit, and sink nodes should report that source metadata to the monitoring system along with the node's own metadata

The details are only defined here for transit. I don't really want to remove "(that was received in the INT-MX header)" since it helps reiterate the context.

I don't really want to move this text somewhere else since I like all the transit node's bit manipulations to be described together.

telemetry/specs/INT.mdk
mickeyspiegel: I numbered the 3 options and added to the next paragraph:

> The expectation is that each INT node implementation will only support a subset, while monitoring systems should support at least 1 and 3.

mickeyspiegel: and I added to 2:

> 2. The domain specific extension data includes the 'source-only' metadata (note that this is free format, so this may require additional implementation effort by the monitoring system)

rsivakolundu: Why not use the Variable Optional DSMdBits Data to carry "source-only" metadata?

telemetry/specs/INT.mdk
rsivakolundu: I concur with that.

telemetry/specs/INT.mdk
mickeyspiegel: @rsivakolundu Are you ok with removing the 2nd option of using "domain specific extension data" to carry MX source-only metadata, leaving the truncated packet and Variable Optional DSMdBits Data options?

jklr (approved):
Thanks. I guess cleanup of the padding option and E/M/HopML will be done in a separate PR.

mickeyspiegel:
I will open a new PR for removing option #2 (padding), HopML, E, M, which will touch Telemetry Report as well as INT.

PR #87, opened 2020-08-20
INT-MX: Generalize domain specific 'source-only' metadata

mickeyspiegel:
Change INT-MX 'source-only' metadata to 'source-inserted' metadata.
When a DS Instruction bit is defined for a domain, the definition
must specify the following properties:
- DS Instruction Mode: 'Export' or 'Source-Inserted'
  (new terminology for behavior already defined in INT v2.1)
- Source-Inserted Metadata Reporting Requirement:
  'All Nodes', 'Sink Node', or 'None'
- Source-Inserted Metadata Mutability:
  'Source-Only' (immutable) or 'Cumulative' (mutable)


jklr:
+ @rsivakolundu

telemetry/specs/INT.mdk
```diff
@@ -1408,25 +1412,48 @@ INT-MX Header:
...
-  they indicate that the metadata is embedded, and that each node including
-  the source, transit, and sink nodes should report that source metadata to the
-  monitoring system along with the node's own metadata. Note that there are
-  two ways to include the 'source-only' metadata in the Telemetry Report,
-  described below.
+  identified by the *Domain Specific ID*. When a bit is defined for a particular
+  domain, in addition to the metadata semantics and syntax, the definition must
+  specify the behavior with respect to the following properties:
+
+    - **DS Instruction Mode:**
```
jklr, line 1432: Below two flag 'bits' (Export, Source-Inserted) are mentioned part of each DS instruction definition. Ppl may think these bits as included in either INT or telemetry headers, but they're not as I understand. Are we proposing these bits part of formal YANG model?
Wonder if we can explain them as 'properties' of each DS instruction that each domain has to decide and specify in some form, e.g., the YANG model.

mickeyspiegel: They are not described as "bits". It says "properties". And absolutely they should be included in the YANG model.

rsivakolundu (approved):

PR #88, opened 2020-08-20
INT v2.1: Add Congestion Considerations section

mickeyspiegel:
At the request of IETF experts, in response to the application for a UDP destination port codepoint.


jklr (approved):

jklr:
@mickeyspiegel shall we resolve a conflict and merge?

PR #89, opened 2020-10-09
Telemetry Report: Support packet fragments longer than 1020 bytes

mickeyspiegel:
Some customers do not want to truncate the packet fragments included in
telemetry reports. This commit defines the maximum value of 'Report Length'
0xFF as a special value that indicates a length greater than or equal to
0xFF in 4-byte words. That value indicates that this individual report
extends to the end of the UDP payload, i.e. there are no subsequent
individual reports in this telemetry report.

When coalescing of multiple individual reports into a single telemetry
report is desired, the operator must specify a maximum truncate length less
than or equal to 254 * 4 = 1016 bytes, including metadata and other
individual report fields.


jklr:
@rsivakolundu

rsivakolundu (commented):
Looks good to me.

jklr (approved):

